PR #23, opened 2015-11-20
Gracefully handle empty json field

eiriklied:
This PR comes from a problem we experienced when Unleash was trying to load feature toggles from the `unleash-repo.json`. The JSON parsing itself went well, but when the expected field `features` was not present we would pass `null` into `ToggleCollection` and end up getting a NPE when instantiating `DefaultUnleash`.

Here are two fixes to solve this problem:

1. `ToggleCollection` will now gracefully handle `null` input that would cause it to throw NPE.
1. We now validate that the `features` field is actually present after parsing JSON. If not, `IllegalStateException` is thrown and will be handled just like other cases of invalid JSON.


ivarconr:
Looks good!

PR #24, opened 2016-02-01
Made example work by adding /features to URI path

emilva:
Also updated URI to the running public demo instance of unleash-server


sveinelo:
Thank you!

PR #26, opened 2016-08-02
Make FakeUnleash differ "not set" from disabled

sagen:
In FakeUnleash, you can't really set a toggle to be disabled, as it just unsets the enabled feature. This meant that unleash.enabled("feature", true) would incorrectly return true for a disabled feature.


ivarconr:
Yes indeed you have a solid point. Thanks!

PR #30, opened 2016-11-09
Implement client registration and metrics.

ivarconr:
Adds implementation where the client register at startup
and sends metrics to the server regularly.

Needs one more thing:
- An option to disable metrics.

@sveisvei


sveisvei:
Lgtm

Issue #29, opened 2016-11-05
Extend the unleash interface with context-data


ivarconr:
In the current version of Unleash strategies needs to get their context data via special provides which typically is implemented as some magic-data available at the current request. This works well in most java-based web-applications. 

However doing toggles in a standalone web-application this becomes much harder as you typically do not have a user bound to the current request. In this scenario you will typically have the context-data available runtime, at the same location you want to verify if a toggle should be on or of. 

It would thus be beneficial if the client also could inject context data at runtime. The unleash API would then be overloaded with a new method, the the user can inject the required context-data. 

``` java
unleash.isEnabled("app.someToggle", UnleashConext.of("k1", "v1", "k2", "v2"));
```

Unleash will then in turn inject the context-data to the strategy which would be extended to:

``` java
public interface Strategy {
    String getName();

    boolean isEnabled(Map<String, String> parameters);

    boolean isEnabled(Map<String, String> parameters, UnleashConext context);
}
```

(\* if we move the client to java8 we could also provide a default implementation of the new method to avoid breaking the API)

This is also the approach used for the [unleash-client-node](https://github.com/finn-no/unleash-client-node).

ivarconr:
(from https://github.com/finn-no/unleash/issues/99)

ivarconr:
After years of usage i finn we see that a UnleashUser with the following properties would make it simple to standardize a lot of strategies (all fields are of course optional):

```
UnleashUser::
 identifier: String,
 remoteAddress: String,
 sessionId: String,
 properties: Map<String, String>
```

The UnleashUser should be part of the `UnleashContext` and it should be possible to provide the context via a request-bound provider (e.g. in a request scoped spring bean, or similar). It should also be possible to send in the `UnleashContext` (with an unleashUser on it) as part of the "isEnabled" call. This is very convenient in non-thread based environment where the user/request details is available just in place.

ivarconr:
Suggestion: 

should we change the `properties` field to be a just a list of flags instead? The only use case we have for it at the moment us to flag a user as a beta-user. This feels a bit cumbersome in a key,value structure. 

Aka:

```UnleashContext::
 userId: String,
 sessionId: String,
 remoteAddress: String,
 flags: Array<String>
```

What do you think @sveisvei  and @vsandvold ?

vsandvold:
[...]

PR #33, opened 2017-02-05
Add support for UnleashContext #29

ivarconr:
The idea is to simplify common strategies via a shared unelash context. The context should be possible to provide just in time, or via a request bound probider. 

**Just in time example:**

```java
UnleashContext context = UnleashContext.builder().userId("user@mail.com").build();
unleash.isEnabled("someToggle", unleashContext);
```
The main benefit of just in time is that it is more explicit. It is also makes it easier when ThreadLocal is not bound to the current request. 

**Provider example:**
```java
UnleashContextProvider provider = new CustomUnleashContextProvider();
UnleashConfig config = UnleashConfig.builder().contextProvider(provider);
Unleash unleash = new DefaultUnleash(config):

// context will be provided to the strategy impl via the provided contextProvider. 
unleash.isEnabled("someToggle");
```

this is still WIP!

todo:

- [x] Implement UserContext
- [x] Implement common strategies: 
  - [x] UserWithIdStrategy
  - [x] GradualRolloutRandom Strategy
  - [x] GradualRolloutUserWithId Strategy
  - [x] GradualRolloutSessionId Strategy
  - [x] RemoteAddress Strategy
  - [x] ApplicationHostname Strategy
- [x] Add support for providing unleashContext (via config and inside UnleashDefault)
- [x] Unleash should add all "default" strategies automatically. 
- [x] Write some documentation on how to provide the UnleashContext ("just in time" vs provider)
- [x] Do some testing before merging.

### DIFF
--- a/src/main/java/no/finn/unleash/strategy/GradualRolloutRandomStrategy.java
+++ b/src/main/java/no/finn/unleash/strategy/GradualRolloutRandomStrategy.java
@@ -7,7 +7,7 @@ import java.util.Random;
 import no.finn.unleash.strategy.Strategy;
 
 public final class GradualRolloutRandomStrategy implements Strategy {
-
+    protected static final String PERCENTAGE = "percentage";
     private static final String STRATEGY_NAME = "gradualRolloutRandom";
 
     private final Random random;
@@ -27,10 +27,8 @@ public final class GradualRolloutRandomStrategy implements Strategy {
 
     @Override
     public boolean isEnabled(final Map<String, String> parameters) {
-        return Optional.ofNullable(parameters.get("percentage"))
-                .filter(percentageStr -> percentageStr.matches("^100|0|[1-9][0-9]?$"))
-                .map(Integer::parseInt)
-                .map(percentage -> percentage >= random.nextInt(100) + 1)
-                .orElse(false);
+        int percentage = StrategyUtils.getPercentage(parameters.get(PERCENTAGE));
+        int randomNumber = random.nextInt(100) + 1;
+        return percentage >= randomNumber;
     }
 }
--- a/src/main/java/no/finn/unleash/strategy/GradualRolloutRandomStrategy.java
+++ b/src/main/java/no/finn/unleash/strategy/GradualRolloutRandomStrategy.java
@@ -1,0 +1,36 @@
+package no.finn.unleash.strategy;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.Random;
+
+import no.finn.unleash.strategy.Strategy;
+
+public final class GradualRolloutRandomStrategy implements Strategy {
+
+    private static final String STRATEGY_NAME = "gradualRolloutRandom";
+
+    private final Random random;
+
+    public GradualRolloutRandomStrategy() {
+        random = new Random();
+    }
+
+    protected GradualRolloutRandomStrategy(long seed) {
+        random = new Random(seed);
+    }
+
+    @Override
+    public String getName() {
+        return STRATEGY_NAME;
+    }
+
+    @Override
+    public boolean isEnabled(final Map<String, String> parameters) {
+        return Optional.ofNullable(parameters.get("percentage"))
+                .filter(percentageStr -> percentageStr.matches("^100|0|[1-9][0-9]?$"))
+                .map(Integer::parseInt)
+                .map(percentage -> percentage >= random.nextInt(100) + 1)
+                .orElse(false);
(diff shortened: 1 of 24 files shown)


src/main/java/no/finn/unleash/strategy/UserWithIdStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public boolean isEnabled(Map<String, String> parameters) {
+        return false;
+    }
+
+    @Override
+    public boolean isEnabled(Map<String, String> parameters, UnleashContext unleashContext) {
+        return unleashContext.getUserId()
+                           .map(currentUserId -> Optional.ofNullable(parameters.get(PARAM))
```
vsandvold: Almost trivial, but util methods for parsing different parameter types would be nice.

src/main/java/no/finn/unleash/strategy/GradualRolloutRandomStrategy.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public String getName() {
+        return STRATEGY_NAME;
+    }
+
+    @Override
+    public boolean isEnabled(final Map<String, String> parameters) {
+        return Optional.ofNullable(parameters.get("percentage"))
+                .filter(percentageStr -> percentageStr.matches("^100|0|[1-9][0-9]?$"))
```
vsandvold: `StrategyUtils.getPercentage(parameters.get(PERCENTAGE))`?

ivarconr: yes, thanks!

src/main/java/no/finn/unleash/strategy/StrategyUtils.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Takes a numeric string value and converts it to a integer between 0 and 100.
+     *
+     * returns 0 if the string is not numeric.
+     *
+     * @param percentage - A numeric string value
+     * @return a integer between 0 and 100
+     */
+    public static int getPercentage(String percentage) {
+        if (isNotEmpty(percentage) && isNumeric(percentage)) {
```
vsandvold: Should it handle `10%` as well?

src/test/java/no/finn/unleash/strategy/GradualRolloutRandomStrategyTest.java
```diff
@@ -0,0 +1,100 @@
...
+        assertFalse(enabled);
+    }
+
+    @Test
+    public void should_never_be_enabled_when_0_percent() {
+        final Map<String, String> parameters = new HashMap<String, String>() {{
+            put("percentage", "0");
+        }};
+
+        for (int i = 0; i < 1000; i++) {
```
vsandvold: Wow, one thousand percent :-)

src/main/java/no/finn/unleash/strategy/StrategyUtils.java
ivarconr: Not to important after the UI got input types. Percentage input looks like this:
image

src/test/java/no/finn/unleash/strategy/GradualRolloutRandomStrategyTest.java
ivarconr: typo :writing_hand:

ivarconr:
Question: 

Should we override the `DefaultUnleash` constructor to also accept a "UnleashContextProvide" or should we add it as an option to the generic UnleashConfig object?

src/main/java/no/finn/unleash/strategy/StrategyUtils.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Takes to string inputs concat them, produce a hashCode and return a normalized value between 0 and 100;
+     *
+     * @param identifier
+     * @param groupId
+     * @return
+     */
+    public static int getNormalizedNumber(String identifier, String groupId) {
+        int hashCode = Math.abs((groupId + ':' + identifier).hashCode());
```
vsandvold: I got this warning once upon a time http://stackoverflow.com/questions/23416264/bad-attempt-to-compute-absolute-value

Can be changed to something like:
```
final int hashCode = (identifier + groupId).hashCode();
return 1 + Math.abs(hashCode % ONE_HUNDRED);
```

PR #35, opened 2017-06-15
Add support for custom HTTP headers in API calls

andereld:
This commit adds a map of custom HTTP headers to UnleashConfig, and
includes these headers in calls to the Unleash server. This may be used
to, e.g., add an `Authorization` field if the server is behind a wrapper
or proxy which requires authentication.

See Unleash/unleash#222 on GitHub for background discussion.

### DIFF
--- a/src/main/java/no/finn/unleash/util/UnleashConfig.java
+++ b/src/main/java/no/finn/unleash/util/UnleashConfig.java
@@ -1,15 +1,22 @@
 package no.finn.unleash.util;
 
 import java.io.File;
+import java.net.HttpURLConnection;
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
+import java.util.HashMap;
+import java.util.Map;
 
 import no.finn.unleash.UnleashContextProvider;
 
 public class UnleashConfig {
+    protected static final String UNLEASH_APP_NAME_HEADER = "UNLEASH-APPNAME";
+    protected static final String UNLEASH_INSTANCE_ID_HEADER = "UNLEASH-INSTANCEID";
+
     private final URI unleashAPI;
     private final UnleashURLs unleashURLs;
+    private final Map<String, String> customHttpHeaders;
     private final String appName;
     private final String instanceId;
     private final String backupFile;
@@ -20,6 +27,7 @@ public class UnleashConfig {
 
     public UnleashConfig(
             URI unleashAPI,
+            Map<String, String> customHttpHeaders,
             String appName,
             String instanceId,
             String backupFile,
@@ -38,6 +46,7 @@ public class UnleashConfig {
         }
 
         this.unleashAPI = unleashAPI;
+        this.customHttpHeaders = customHttpHeaders;
         this.unleashURLs = new UnleashURLs(unleashAPI);
         this.appName = appName;
         this.instanceId = instanceId;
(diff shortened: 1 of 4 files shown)


src/main/java/no/finn/unleash/util/UnleashConfig.java
```diff
@@ -88,8 +101,17 @@ public UnleashContextProvider getContextProvider() {
         return contextProvider;
     }
 
+    public static void setRequestProperties(HttpURLConnection connection, UnleashConfig config) {
+        connection.setRequestProperty(UNLEASH_APP_NAME_HEADER, config.getAppName());
+        connection.setRequestProperty(UNLEASH_INSTANCE_ID_HEADER, config.getInstanceId());
```
SimenB: Can this set `User-Agent` as well? Since this is getting a small refactor either way 🙂

Ref Unleash/unleash-client-node#39

src/main/java/no/finn/unleash/util/UnleashConfig.java
```diff
@@ -88,8 +101,17 @@ public UnleashContextProvider getContextProvider() {
...
+        for (String name : config.getCustomHttpHeaders().keySet()) {
```
SimenB: Is this module java 8? If so, can use `config.getCustomHttpHeaders().keySet().forEach`. If _not_ java 8, ignore!

src/main/java/no/finn/unleash/util/UnleashConfig.java
```diff
@@ -1,15 +1,22 @@
...
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
+import java.util.HashMap;
+import java.util.Map;
 
 import no.finn.unleash.UnleashContextProvider;
 
 public class UnleashConfig {
+    protected static final String UNLEASH_APP_NAME_HEADER = "UNLEASH-APPNAME";
```
SimenB: can be package local instead of protected, right?

src/main/java/no/finn/unleash/util/UnleashConfig.java
andereld: `maven-compiler-plugin` in the `pom.xml` specifies a source and target of 1.8, so I guess that makes the module Java 8. I'll change it to `forEach`, that is nicer – thanks!

src/main/java/no/finn/unleash/util/UnleashConfig.java
andereld: Yes, I'll change it, thanks.

src/main/java/no/finn/unleash/util/UnleashConfig.java
andereld: Sure. I didn't think to add it as there was no `User-Agent` in the code to begin with, but I'll add it now. Based on the linked commit, I take it the `User-Agent` should be the app name.

ivarconr:
Nice job, even got tests! Thanks =)

ivarconr:
Will release this next week (traveling this weekend, and do not have correct keys set up on my laptop).

ivarconr:
Released as 2.1.2!

andereld:
Nice, thanks!

> `config.getCustomHttpHeaders().forEach(connection::setRequestProperty);`

😄

PR #36, opened 2017-06-25
Add method to get the featureToggle definition

vsandvold:


vsandvold:
It works, preferably without `getFeatureToggleDefinition` in the interface, just as a public method on `DefaultUnleash. Otherwise, we must implement it in `FinnUnleash` aswell.

Another caveat with `FinnUnleash` is that we must cast the reference everywhere we want to use the `startExperiment` method. Maybe we should add it to the interface right away, and provide a default implementation for backward compatibility?

SimenB:
>Maybe we should add it to the interface right away, and provide a default implementation for backward compatibility?

👍

src/main/java/no/finn/unleash/FakeUnleash.java
```diff
@@ -24,6 +24,10 @@ public boolean isEnabled(String toggleName, boolean defaultSetting) {
         }
     }
 
+    public FeatureToggle getFeatureToggleDefinition(String toggleName) {
```
ivarconr: can be dropped as it is not part of the interface!

ivarconr:
> Maybe we should add it to the interface right away, and provide a default implementation for backward compatibility?

I think we should do that as soon as we also can ship a working `abTestStrategy` we are happy with!

src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -92,6 +92,10 @@ public boolean isEnabled(final String toggleName, final UnleashContext context ,
         return enabled;
     }
 
+    public FeatureToggle getFeatureToggleDefinition(String toggleName) {
+        return toggleRepository.getToggle(toggleName);
```
vsandvold: Is this method nullable, in case of a non-existent toggleName?

ivarconr: then it will return null, could preferably be wrapped in to an optional perhaps?

vsandvold:
That could work, I'll fix it tomorrow

Den man. 26. jun. 2017, 21.13 skrev Ivar Conradi Østhus <
notifications@github.com>:

PR #44, opened 2017-10-03
Implement hashing with murmur

ivarconr:
Should be merged in 3.x version of client.


sveisvei:
@ivarconr missing a test?

ivarconr:
Yes!. We should even provide standardized test vectors so we can confirm
normalisation across client implementations.

ivarconr:
see #45

PR #45, opened 2017-11-16
Implement hashing with murmur

alanchristensen:
This is @ivarconr's PR with some tests added to match the node client.

If these look good then I'll work on the go client next.

Node client PR https://github.com/Unleash/unleash-client-node/pull/54


ivarconr:
Thanks @alanchristensen for taking the time to fix this! :smile: :fire:

PR #48, opened 2018-06-12
Expose list of feature names.

chadjosephson-octanner:


chadjosephson-octanner:
This pull request is meant to expose the list of features available, the list of enabled features, and the list of enabled features for a given context.

src/main/java/no/finn/unleash/repository/FeatureToggleRepository.java
```diff
@@ -1,12 +1,14 @@
 package no.finn.unleash.repository;
 
+import no.finn.unleash.UnleashContext;
```
ivarconr: Please remove unused import.

src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -97,6 +98,20 @@ public boolean isEnabled(final String toggleName, final UnleashContext context ,
...
+    }
+
+    @Override
+    public List<String> getEnabledToggleNames() {
+        return getEnabledToggleNames(contextProvider.getContext());
+    }
+
+    @Override
+    public List<String> getEnabledToggleNames(final UnleashContext context) {
+        return toggleRepository.getToggleNames().stream().filter(toggleName -> this.isEnabled(toggleName, context, false)).collect(Collectors.toList());
```
ivarconr: For readability this should be split over multiple lines. Generally I would prefer it like this:

```java
return toggleRepository.getToggleNames().stream()
    .filter(toggleName -> this.isEnabled(toggleName, context, false))
    .collect(Collectors.toList());
```

src/main/java/no/finn/unleash/FakeUnleash.java
```diff
@@ -24,6 +26,21 @@ public boolean isEnabled(String toggleName, boolean defaultSetting) {
...
     }
 
+    @Override
+    public List<String> getToggleNames() {
+        return features.keySet().stream().collect(Collectors.toList());
+    }
+
+    @Override
+    public List<String> getEnabledToggleNames() {
+        return features.keySet().stream().filter(feature -> features.get(feature)).collect(Collectors.toList());
```
ivarconr: Use multiple lines for readability.

src/main/java/no/finn/unleash/repository/FeatureToggleRepository.java
```diff
@@ -50,4 +52,10 @@ private Runnable updateToggles() {
     public FeatureToggle getToggle(String name) {
         return toggleCollection.getToggle(name);
     }
+
+    @Override
+    public List<String> getToggleNames() {
+        return toggleCollection.getFeatures().stream().map(toggle -> toggle.getName()).collect(Collectors.toList());
```
ivarconr: Multiple lines for readability

src/main/java/no/finn/unleash/FakeUnleash.java
```diff
@@ -24,6 +26,21 @@ public boolean isEnabled(String toggleName, boolean defaultSetting) {
...
+    }
+
+    @Override
+    public List<String> getEnabledToggleNames(UnleashContext context) {
+        return getEnabledToggleNames();
+    }
+
```
ivarconr: Missing test cases for the FakeUnleash. It's important to verify the behaviour of the FakeUnleash implementation, as many users relies on it to run their own unit-tests, independent of the unleash infrastructure.

ivarconr:
Thanks for the contribution @chadjosephson-octanner. I can understand the need for getting the status of multiple toggles at once, but it would be nice if you could elaborate a bit on what you intend to use the `getEnabledToggleNames` for. 

However there are a few things we should consider: 
1. Many companies uses Unleash across many applications and services. A typical convention is to prefix toggles to get a logical separation (e.g. `app.toggleNamne`) Should these `getEnabledToggleNames` accept a "name filter", so that the application can query only toggles relevant for that application? 
2. One side-effect of calling `isEnabled` is that the client will count yes/no for the toggle being tested. Have the application used the toggle if it calls `getEnabledToggleNames`? Do we have to omit metrics in that scenario? The intention of metrics in unleash is to keep track of which application is using which toggle and what the feature toggle evaluates to. Another use case for metrics is to detect unused toggles (no clients querying `isEnabled` for them anymore).

chadjosephson-octanner:
Thank you for reviewing the proposed changes @ivarconr.

The use-case these proposed changes addresses relates to how our application stack is structured. We deliver a single page application backed by many micro-service endpoints. The intent is to have the layout microservice deliver a list of feature toggles to the front-end rather than having each microservice query Unleash. 

This could be accomplished by maintaining a list of potential feature toggles in the layout microservice and querying each flag individually.  What I don't like about that solution is the layout microservice would need to know about all potential features for all microservices and the UI. That would create unnecessary dependencies between services.  In my opinion it would be much simpler to ask Unleash to return a list of all enabled feature toggles. 

Let me know if this concept is in-line with Unleash. In the meantime I will address specific things you have brought up in the review. Thanks again for your time.

src/main/java/no/finn/unleash/FakeUnleash.java
chadjosephson-octanner: Thank you for the clarification of what this class is used for. It wasn't immediately obvious to me.

chadjosephson-octanner:
Regarding your two considerations:

1. I would leave using the application name to filter as a question for you. At the moment it wouldn't be of any value to my project.

2. I can see how my changes will cause the metrics to have no value. Knowing which features are no longer being queried is useful information to have and I am not sure how to reconcile existing metrics and my changes.

ivarconr:
Adding the `getToggleNames` is a no-brainer, and we should just do that. 

Given that you get this from the Unleash API, how important is it that the client actually provides you with the `getAllEnabledToggles` implementation?  (It would be fairly simple to have the "stream->filter->collect" impl on your side anyway). 

My concern is that it is not obvious for the users that this will reduce the value of the metrics.

What do you think?

ivarconr:
Another approach could be to add the method `Map<String, boolean> evaluateAllToggles()`, which in my opinion is more explicit, but it does not solve your use case directly.

ivarconr:
And a big thanks for taking time to reach out and suggest improvements. I am always happy when users talk about their needs and find ways to improve the API. 

I will make sure to merge the changes and get a release out asap.

chadjosephson-octanner:
Didn't want you to merge while I was going through the cleanup/squash/force process on my fork.

chadjosephson-octanner:
I have reverted all of the `getEnabledToggleNames` code and left it with `getFeatureToggleNames`.

Thanks again Ivar.

ivarconr:
..and released as 3.1.0 :smile:

PR #50, opened 2018-06-13
Introduced UNAVAILABLE as possible status in FeatureToggleResponse

### LINKED ISSUE #49: Unable to detect failures in HttpToggleFetcher
The current API of Unleash/HttpToggleFetcher makes it difficult to detect if HttpToggleFetcher is actually able to fetch toggles or if it experiences for example network failures, authentication/authorization failures or simply 404s because because somebody shut down or moved the unleash server.

It would be nice if applications are able to observe such problems through the API.

jan-berge-ommedal:
closes #49

### DIFF
--- a/src/main/java/no/finn/unleash/DefaultUnleash.java
+++ b/src/main/java/no/finn/unleash/DefaultUnleash.java
@@ -56,10 +56,6 @@ public final class DefaultUnleash implements Unleash {
     }
 
     public DefaultUnleash(UnleashConfig unleashConfig, ToggleRepository toggleRepository, Strategy... strategies) {
-        this(unleashConfig, toggleRepository, unleashScheduledExecutor, strategies);
-    }
-
-    public DefaultUnleash(UnleashConfig unleashConfig, ToggleRepository toggleRepository, UnleashScheduledExecutor unleashScheduledExecutor, Strategy... strategies) {
         this.toggleRepository = toggleRepository;
         this.strategyMap = buildStrategyMap(strategies);
         this.contextProvider = unleashConfig.getContextProvider();
--- a/src/main/java/no/finn/unleash/DefaultUnleash.java
+++ b/src/main/java/no/finn/unleash/DefaultUnleash.java
@@ -56,6 +56,10 @@ public final class DefaultUnleash implements Unleash {
     }
 
     public DefaultUnleash(UnleashConfig unleashConfig, ToggleRepository toggleRepository, Strategy... strategies) {
+        this(unleashConfig, toggleRepository, unleashScheduledExecutor, strategies);
+    }
+
+    public DefaultUnleash(UnleashConfig unleashConfig, ToggleRepository toggleRepository, UnleashScheduledExecutor unleashScheduledExecutor, Strategy... strategies) {
         this.toggleRepository = toggleRepository;
         this.strategyMap = buildStrategyMap(strategies);
         this.contextProvider = unleashConfig.getContextProvider();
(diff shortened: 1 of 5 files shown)


src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -56,6 +56,10 @@ public DefaultUnleash(UnleashConfig unleashConfig, Strategy... strategies) {
     }
 
     public DefaultUnleash(UnleashConfig unleashConfig, ToggleRepository toggleRepository, Strategy... strategies) {
+        this(unleashConfig, toggleRepository, unleashScheduledExecutor, strategies);
+    }
+
+    public DefaultUnleash(UnleashConfig unleashConfig, ToggleRepository toggleRepository, UnleashScheduledExecutor unleashScheduledExecutor, Strategy... strategies) {
```
ivarconr: How is overloading the constructor with providing your own `UnleashScheduledExecutor` related to the issue this PR tries to solve?

src/test/java/no/finn/unleash/util/UnleashConfigTest.java
```diff
@@ -49,7 +50,7 @@ public void should_generate_backupfile() {
                 .build();
 
         assertThat(config.getAppName(), is("my-app"));
-        assertThat(config.getBackupFile(), is(System.getProperty("java.io.tmpdir") + "/unleash-my-app-repo.json"));
+        assertThat(config.getBackupFile(), is(System.getProperty("java.io.tmpdir") + File.separatorChar + "unleash-my-app-repo.json"));
```
ivarconr: :+1:

ivarconr:
I agree to the issue and I am happy to merge this (without the overloaded constructor, that must be solved as another PR if needed). 

However; 
I still don't feel the purposed changes makes it super-easy for users of the `unleash-client` to detect these problems. It could be that the `unleash-client` should provide a way for users to register themselves (observer pattern, event-emitter, or similar) and get notified on such errors. This is probably something that should be solved as a separate issue. 

What do you think? Does it make sense?

src/main/java/no/finn/unleash/DefaultUnleash.java
jan-berge-ommedal: Its not directly related, but it did allow me to inject a mocked UnleashScheduledExecutor while testing this to prevent schedulation in the background. 

Also, I think it would be a nice change in general to be able to supply the same custom UnleashScheduledExecutor to DefaultUnleash as to FeatureToggleRepository in this manner:

```

UnleashScheduledExecutor unleashScheduledExecutor = new MyCustomUnleashScheduledExecutor();
        FeatureToggleRepository featureToggleRepository = new FeatureToggleRepository(
                unleashConfig,
                unleashScheduledExecutor,
                new HttpToggleFetcher(unleashConfig),
                new ToggleBackupHandlerFile(unleashConfig)
        );
        Unleash unleash = new DefaultUnleash(unleashConfig, featureToggleRepository, unleashScheduledExecutor);

```

jan-berge-ommedal:
I sent you a pm on slack that you seem to have missed, but yes - I agree. It's still not super-easy for users of the high-level API to detect such problems. At the moment I am wrapping unleash in a feature-toggle-module that will be used across several applications, and in this case it's ok to access the low-level API.

I like your idea of being able to listen events a lot and I prefer it to the alternative of extending the API to be able to pull this information. I would be happy to contribute to this if you would want that.

PR #52, opened 2018-06-20
Upgrade to Junit5

ivarconr:
Todo:

- [x] Migrate all tests to use junit5 and not junit4
- [x] Fix coverage. Might need to switch to Jacoco

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,23 +59,52 @@
         
         <!-- Test dependencies -->
         <dependency>
-            <groupId>junit</groupId>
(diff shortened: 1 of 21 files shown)


ivarconr:
Coverage seems to be working just fine. Will look at Jacoco as a separate task.

ivarconr:
Number of tests went from 131 to 122. This is correct because of a change in the Parameterized test `RemoteAddressStrategyTest`, where we now only execute the should_hava_a_name once, and not once per parameter.

PR #56, opened 2018-06-29
Add client-specification tests

ivarconr:
Implement support for running the specifications defined in https://github.com/Unleash/client-specification


src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -82,6 +82,8 @@ public boolean isEnabled(final String toggleName, final UnleashContext context ,
             enabled = defaultSetting;
         } else if(!featureToggle.isEnabled()) {
             enabled = false;
+        } else if(featureToggle.getStrategies().size() == 0) {
```
ivarconr: The client did not handle the case of empty strategies correctly. Discovered with the help of the client-specifications :+1:

Issue #57, opened 2018-08-28
Synchronous call to pull toggles when instantiating the unleash client


ryanm35:
Right now it looks like when the unleash client is instantiated a background thread is setup to query the toggles from the server. I am running into an issue where a call to `getFeatureToggleNames()` is returning an empty list because the call is made too soon after the instantiation of the client and no backup file existed. 

Can the instantiation make an synchronous call to update first and then setup the background thread?

ivarconr:
Released as part of `unleash-client-java:3.1.1`

PR #58, opened 2018-08-28
feat(initalization): Add feature to try one toggle-update at initialization

ivarconr:


ivarconr:
@ryanm35 will this solve #57 for you? 

It will allow you to define the option `syncFirstFetch` which will instruct the unleash client to perform an initial fetch synchronously, at instantiation before returning. 

It can however fail, this will not throw any exceptions, but hopefully recover when `unleash-api` is available again

src/main/java/no/finn/unleash/util/UnleashConfig.java
```diff
@@ -26,6 +26,7 @@
     private final long sendMetricsInterval;
     private final boolean disableMetrics;
     private final UnleashContextProvider contextProvider;
+    private final boolean syncFirstFetch;
```
ivarconr: I don't fancy this name for this option. Suggestions welcomed!

ivarconr: Thanks I like `syncFetchOnInit`!

ryanm35:
Thanks @ivarconr, this config option would resolve my issue.

src/main/java/no/finn/unleash/util/UnleashConfig.java
SimenB: is `sync` for `synchronous` or `synchronizing`? Not a big fan of abbreviations in code, we all have editors that autocompletes for us anyways 🙂

ryanm35:
I think this feature could be useful for other clients as well. I am going to have a similar problem with with the golang client.

src/main/java/no/finn/unleash/util/UnleashConfig.java
ivarconr: So `synchronousFetchOnInitialization` then?

ivarconr:
@ryanm35 this should be opened as issues per client-impl as the solution will have to follow platform conventions. 

> E.g. the node client is already an event emitter and the events would be a way for the user to block on the initial fetch.

PR #61, opened 2019-01-24
feat: Implement support for variants

ivarconr:


src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -95,6 +100,51 @@ public boolean isEnabled(final String toggleName, final UnleashContext context ,
         return enabled;
     }
 
+    @Override
+    public Variant getVariant(String toggleName, UnleashContext context) {
+        return getVariant(toggleName, context, "");
+    }
+
+    @Override
+    public Variant getVariant(String toggleName, UnleashContext context, String defaultPayload) {
```
ivarconr: The user should be allowed to inject a default `Variant`, not just the payload.

src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -95,6 +100,51 @@ public boolean isEnabled(final String toggleName, final UnleashContext context ,
...
+        final  FeatureToggle featureToggle = toggleRepository.getToggle(toggleName);
+        final boolean isEnabled = isEnabled(toggleName, context, false, featureToggle);
+
+        if(isEnabled) {
+            return selectVariant(featureToggle, context)
+                .map(variantDefinition -> new Variant(variantDefinition.getName(), variantDefinition.getPayload(), isEnabled))
+                .orElseGet(() -> new Variant("default", defaultPayload, isEnabled));
```
ivarconr: The fallback needs to be stream-lined across all client implementations. 

In the code we should also hard-code a fallback-variant, no need to create  a new one every time. 
`Variant fallback = new Variant("disabled", false)`

src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -95,6 +100,51 @@ public boolean isEnabled(final String toggleName, final UnleashContext context ,
...
+
+        int num = 0;
+        for (VariantDefinition definition: featureToggle.getVariants()){
+            if(score + num >= definition.getWeight()){
+                return Optional.of(definition);
+            }
+            num += definition.getWeight();
+        }
+
+        //Should not happen
```
ivarconr: What should not happen?

stalet: Edge case when we are done with the loop. If the result of the modulus operator is higher than max. This is now treated as return empty as we talked about in the discussion this morning.

src/main/java/no/finn/unleash/DefaultUnleash.java
stalet: The payload default payload may differ. We should return the specified default payload on "disabled" right ?

src/main/java/no/finn/unleash/DefaultUnleash.java
stalet: Fixed.

src/main/java/no/finn/unleash/DefaultUnleash.java
stalet: Fixed.

sveisvei:
💯

ivarconr:
only merged to branch so that i can add the override support

PR #64, opened 2019-02-06
fix: instanceId cannot be null.

ivarconr:


ivarconr:
@ben-bourdin451 Thanks for reporting. This fix will make the scenario a bit better. 

Next step should be to look in to pushing events from the client, making it easier to use in client applications.

Issue #51, opened 2018-06-19
Publish events


ivarconr:
The client should provide a way for users to subscribe on interesting events. Preferably without introducing third party dependencies. 

Examples:

- `ready` - Unleash client has fetched initial state from unleash-server is able to serve request
- `update` - Unleash client got updated toggle configuration from server.  
- `error` - Unleash client failed fetching data from unleash-server (might want to set some retry limit?).

ivarconr:
I imagine some way of users to register them as subscribers to receive callbacks for events. We should use [WeakReference](https://docs.oracle.com/javase/7/docs/api/java/lang/ref/WeakReference.html) to store the internal subscriber object, so that stuff can be garbage-collected if the only thing holding a reference is the list of subscribers.

ivarconr:
Would it be doable to follow the event-emitter pattern from node.js?

```java
DefaultUnleash unleash = new DefaultUnleash();
unleash.on("ready", (event) -> {
   System.out.prrintln(event);
});
```

ivarconr:
The observable pattern could be one way to go, I have tested a small implementation here: 
https://github.com/Unleash/unleash-client-java/compare/events/observable?expand=1

PR #67, opened 2019-02-08
feat: subscriber api

jan-berge-ommedal:
- subscribe to all or some events by implementing an UnleashSubscriber
 - improved support for supplying a custom UnleashScheduledExecutor through UnleashConfig
 - default UnleashScheduledExecutor is instantiated lazily
 - isolate periodical task from one-off tasks
 - adding assert4j for more fluent assertions
 - bugfix in FeatureToggleRepository.updateToggles() - do not log error
when status=NOT_CHANGED

targets issue #51

### DIFF
--- a/src/main/java/no/finn/unleash/event/UnleashSubscriber.java
+++ b/src/main/java/no/finn/unleash/event/UnleashSubscriber.java
@@ -1,0 +1,25 @@
+package no.finn.unleash.event;
+
+import no.finn.unleash.UnleashException;
+import no.finn.unleash.metric.ClientMetrics;
+import no.finn.unleash.metric.ClientRegistration;
+import no.finn.unleash.repository.FeatureToggleResponse;
+import no.finn.unleash.repository.ToggleCollection;
+import org.apache.logging.log4j.LogManager;
+
+public interface UnleashSubscriber {
+
+    default void onError(UnleashException unleashException) {
+        LogManager.getLogger(UnleashSubscriber.class).warn(unleashException.getMessage(), unleashException);
+    }
+
+    default void on(UnleashEvent unleashEvent) { }
+    default void onReady(UnleashReady unleashReady) { }
+    default void toggleEvaluated(ToggleEvaluated toggleEvaluated) { }
+    default void togglesFetched(FeatureToggleResponse toggleResponse) { }
+    default void clientMetrics(ClientMetrics clientMetrics) { }
+    default void clientRegistered(ClientRegistration clientRegistration) { }
+    default void togglesBackedUp(ToggleCollection toggleCollection) { }
+    default void toggleBackupRestored(ToggleCollection toggleCollection) { }
+
+}
--- a/src/main/java/no/finn/unleash/metric/UnleashMetricsSender.java
+++ b/src/main/java/no/finn/unleash/metric/UnleashMetricsSender.java
@@ -2,6 +2,7 @@ package no.finn.unleash.metric;
 
 import com.google.gson.*;
 import no.finn.unleash.UnleashException;
+import no.finn.unleash.event.EventDispatcher;
 import no.finn.unleash.util.UnleashConfig;
 import no.finn.unleash.util.UnleashURLs;
 import org.apache.logging.log4j.LogManager;
@@ -18,16 +19,17 @@ import java.time.ZoneOffset;
 import static java.time.format.DateTimeFormatter.ISO_INSTANT;
 
 public class UnleashMetricsSender {
-    private static final Logger LOG = LogManager.getLogger(UnleashMetricsSender.class);
     private static final int CONNECT_TIMEOUT = 1000;
 
     private final Gson gson;
+    private final EventDispatcher eventDispatcher;
     private UnleashConfig unleashConfig;
     private final URL clientRegistrationURL;
     private final URL clientMetricsURL;
 
     public UnleashMetricsSender(UnleashConfig unleashConfig) {
         this.unleashConfig = unleashConfig;
+        this.eventDispatcher = new EventDispatcher(unleashConfig);
         UnleashURLs urls =  unleashConfig.getUnleashURLs();
         this.clientMetricsURL = urls.getClientMetricsURL();
         this.clientRegistrationURL = urls.getClientRegisterURL();
@@ -49,8 +51,9 @@ public class UnleashMetricsSender {
         if(!unleashConfig.isDisableMetrics()) {
             try {
                 post(clientRegistrationURL, registration);
+                eventDispatcher.dispatch(registration);
             } catch(UnleashException ex) {
-                LOG.warn("failed to register client", ex);
+                eventDispatcher.dispatch(ex);
             }
         }
     }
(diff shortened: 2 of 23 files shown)


src/main/java/no/finn/unleash/repository/FeatureToggleRepository.java
```diff
@@ -41,13 +58,18 @@ private Runnable updateToggles() {
...
+                eventDispatcher.dispatch(response);
+                FeatureToggleResponse.Status status = response.getStatus();
+                if (status == FeatureToggleResponse.Status.CHANGED) {
                     toggleCollection = response.getToggleCollection();
                     toggleBackupHandler.write(response.getToggleCollection());
-                } else {
-                    LOG.warn("Error fetching toggles from Unleash API (StatusCode: {})", response.getHttpStatusCode());
+                } else if (status == FeatureToggleResponse.Status.UNAVAILABLE){
+                    String errorMessage = "Error fetching toggles from Unleash API - StatusCode: " + response.getHttpStatusCode();
+                    eventDispatcher.dispatch(new UnleashException(errorMessage, null));
```
ivarconr: Here we might end up dispatching twice for same event, both in line 61 and here.

jan-berge-ommedal: My thinking here was that this case is interesting for subscribers reacting to onError(), instead of having to check the status of FeatureToggleResponse. Alternatively, onError() could be called conditionally from FeatureToggleResponse.publishTo() if that's better. Or I could remove it altogether. What do you prefer?

src/main/java/no/finn/unleash/metric/UnleashMetricsSender.java
```diff
@@ -87,6 +94,7 @@ private int post(URL url, Object o) throws UnleashException {
 
             connection.connect();
 
+            // TODO should probably check response code to detect errors?
```
jan-berge-ommedal: should I do something about this?

src/main/java/no/finn/unleash/util/UnleashConfig.java
```diff
@@ -227,7 +265,10 @@ public UnleashConfig build() {
                     sendMetricsInterval,
                     disableMetrics,
                     contextProvider,
-                    synchronousFetchOnInitialisation);
+                    synchronousFetchOnInitialisation,
+                    Optional.ofNullable(scheduledExecutor).orElseGet(UnleashScheduledExecutorImpl::getInstance),
+                    Optional.ofNullable(unleashSubscriber).orElseGet(NoOpSubscriber::new)
```
jan-berge-ommedal: Is no-op the right default, or should it perhaps be something more substantial like Log4JSubscriber? We could put the full responsibility of handling exceptions on the subscriber so that log-statements like https://github.com/Unleash/unleash-client-java/pull/67/files#diff-7fc449c201ef2331c2e8474f2d9ace6eL51 could be removed...

src/main/java/no/finn/unleash/repository/FeatureToggleRepository.java
ivarconr: I see your point, but it might be the modelling here that is a bit off. 

I don't think that the user cares if it is a error on the HTTP layer or if it some kind of IO-expection. Either there is a response (success) or it's not (error). Is there a use-case where the user would like to get a `FeatureToggleResponse` if it was not successful (instead of an error)? 


I would prefer that `onError` was called conditionally from `FeatureToggleResponse.publishTo()`.

src/main/java/no/finn/unleash/util/UnleashConfig.java
ivarconr: yes, I would prefer that we use subscriber for the logs.

The only thing that worries me then is that worries me then is that you end up losing logs if the user decide to register a custom subscriber and only cares about one of the events (e.g. the metrics). 

Maybe this can to be a separate PR?

ivarconr:
All in all great work! Could you make sure that the first line in the commit use semantic syntax, for example `feat: Subscriber API`

src/main/java/no/finn/unleash/event/UnleashSubscriber.java
```diff
@@ -0,0 +1,21 @@
...
+
+    default void on(UnleashEvent unleashEvent) { }
+    default void onError(UnleashException unleashException) { }
+    default void toggleEvaluated(ToggleEvaluated toggleEvaluated) { }
+    default void togglesFetched(FeatureToggleResponse toggleResponse) { }
+    default void clientMetrics(ClientMetrics clientMetrics) { }
+    default void clientRegistered(ClientRegistration clientRegistration) { }
+    default void togglesBackedUp(ToggleCollection toggleCollection) { }
+    default void toggleBackupRestored(ToggleCollection toggleCollection) { }
+
```
ivarconr: I was thinking that there is one event missing here: `onReady` which should be triggered as soon as the client has successfully fetched from the server.

src/main/java/no/finn/unleash/util/UnleashConfig.java
ivarconr: ...or maybe the default implementation in the interface can perform logging on `onError`?

ivarconr: and maybe we should provide a `DebugSubscriber` people can use to debug Unleash? I see that can be useful from time to time.

jan-berge-ommedal: I like the idea of letting the default UnleashSubscriber.onError() log errors and will refactor towards that. 

Also, the current Log4JSubscriber is already pretty noisy and could probably be used to debug unleash. But i'll add the possibility to adjust the levels directly on the instance to make it more usable.

src/main/java/no/finn/unleash/repository/FeatureToggleRepository.java
jan-berge-ommedal: If the client receives 4xx- or 5xx-responses from the unleash server or some proxy server, the status of FeatureToggleResponse will be `UNAVAILABLE` and no exception will be thrown. The simplest way I can think of to achieve this is to take down the unleash server behind a loadbalancer which then respons with 404s or 503s. 

We had an incident in production where a misbehaving client managed to overwhelm the unleash server (actually the postgres database). Since we had tied alerts to the `UNAVAILABLE`-status, the problem was highlighted quickly to us in what turned out to be almost a pleasing experience, displaying the self-healing properties of both unleash and our infrastructure. 

I'll refactor the `else if`, taking your comments below about logging into account!

src/main/java/no/finn/unleash/event/UnleashSubscriber.java
jan-berge-ommedal: Like this?
https://github.com/Unleash/unleash-client-java/pull/67/files#diff-7fc449c201ef2331c2e8474f2d9ace6eR68
https://github.com/Unleash/unleash-client-java/pull/67/files#diff-c1feded3c37b2e4f9d387af14ea0bf3cR17

ivarconr:
Sorry been a bit occupied last week. What's left to make this one mergable?

jan-berge-ommedal:
> Sorry been a bit occupied last week. What's left to make this one mergable?

There was a conflict with #69 since both PRs actually fixes the same bug. Since the error logging now is performed in the subscriber, this PR effectively negates #69. Oh well.

As far as I can see, this PR is now ready unless you want me to rebase this https://github.com/Unleash/unleash-client-java/pull/67/commits/063e8cba488cb29cf93cc5c80d7e11f242ee68a9

ivarconr:
Please revise ;)

I think you should be allowed to merge it yourself ;)

Issue #68, opened 2019-02-18
Error fetching toggles from Unleash API (StatusCode: 304)


alxsimo:
I have an Unleash server v3.2.1 and java client v3.2.0 on a SpringBoot 2 backend API.

Seems to work well and use the updated value on the backend API but in the console I have this warrning/error every second:

```
2019-02-18 15:26:20.592  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
2019-02-18 15:26:21.865  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
2019-02-18 15:26:22.745  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
2019-02-18 15:26:23.789  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
2019-02-18 15:26:23.842  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
2019-02-18 15:26:24.375  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
2019-02-18 15:26:25.008  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
2019-02-18 15:26:26.282  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
2019-02-18 15:26:26.990  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
2019-02-18 15:26:27.315  WARN 7514 --- [sh-api-executor] n.f.u.r.FeatureToggleRepository          : Error fetching toggles from Unleash API (StatusCode: 304)
[...]

PR #69, opened 2019-02-19
Changed logging HTTP status codes

alxsimo:
This PR fixes the issue discussed in #68. 
Previously we're logged also _good_ status codes.


ivarconr:
thanks :100:

ivarconr:
released as 3.2.1

alxsimo:
Thanks!! :)

PR #74, opened 2019-03-21
fix: Thread safety regarding metrics

aleksabl:
Fixed a couple of issues regarding metrics when
`MetricsBucket.registerCount` is called concurrently

### DIFF
--- a/src/main/java/no/finn/unleash/metric/ToggleCount.java
+++ b/src/main/java/no/finn/unleash/metric/ToggleCount.java
@@ -1,41 +1,43 @@
 package no.finn.unleash.metric;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLong;
 
 class ToggleCount {
-    private long yes;
-    private long no;
-    private Map<String, Long> variants;
+    private final AtomicLong yes;
+    private final AtomicLong no;
+    private final ConcurrentMap<String, AtomicLong> variants;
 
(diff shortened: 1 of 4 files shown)


src/main/java/no/finn/unleash/metric/ToggleCount.java
```diff
@@ -1,41 +1,43 @@
...
+        this.yes = new AtomicLong(0);
+        this.no = new AtomicLong(0);
+        this.variants = new ConcurrentHashMap<>();
     }
 
     public void register(boolean active) {
-        if(active) {
-            yes++;
+        if (active) {
+            yes.incrementAndGet();
```
ivarconr: wow, and I really remember now that `++` is not an atomic operation.

ivarconr:
Thanks, it's nice to get the metric counting thread-safe!

Before merging this I would like to do some performance evaluation. 

Do you have any thoughts on performance @aleksabl ?

aleksabl:
If you create micro benchmark I'm sure you'll find that the code will be a bit slower than the version that only works in a single thread environment. But in the grand scheme of things, I think the performance implications are negible.

ivarconr:
:+1:  I'm sure it probably does not matter, but nice to now what we are talking 1ms vs. 0.003ms

aleksabl:
I'm not sure how much slower it is, but we're definetly not taking about 1ms (or even 0.1ms) here

ivarconr:
I have tested the change (on my machine) it it seems do add about 20ns. I think this is fine and worth it as the accuracy of metrics will be increased. 

If we later want it to be faster we could simply just add results to a list, and perform the counting as part of metrics pusher.

PR #78, opened 2019-06-17
Add support for SLF4J and Make dependency on Log4j 2.x optional

joschi:
A client library, such as unleash-client-java, shouldn't force users to use a specific logging library.

[SLF4J](https://www.slf4j.org/) allows users to select from a wide variety of logging libraries, such as Log4j 1.x, Log4j 2.x, Logback, java.util.logging, and others.


ivarconr:
My thinking has been that we only depend on `log4j-api` and the user can add any bridge she likes (e.g. `log4j-slf4j-impl`). 

Can you enlighten me why it would be better to depend on `slf4j-api`? Does it not exist bridges for log4j2 to most logging frameworks today?

ivarconr:
Side note:

After introducing [subscriber-api](https://github.com/Unleash/unleash-client-java#subscriber-api) I have been thinking we might actually remove the built in logging, and provide examples (or even separate implementation) and remove logging dependency all together.

joschi:
> My thinking has been that we only depend on `log4j-api` and the user can add any bridge she likes (e.g. `log4j-slf4j-impl`).

[log4j-api](https://logging.apache.org/log4j/2.x/log4j-api/index.html) forces users of this library to use Log4j 2.x in their application. There are no other implementations against this API.

[log4j-slf4j-impl](https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/index.html) on the other hand is a Log4j 2.x implementation against the SLF4J API. When using SLF4J, users can decide themselves, whether they want to use java.util.logging, Logback, Log4j, Log4j 2.x, or Apache Commons Logging (which itself is also a wrapper around logging frameworks) in their application.

> Does it not exist bridges for log4j2 to most logging frameworks today?

Only for [Log4j 1.2](https://logging.apache.org/log4j/2.x/log4j-1.2-api/index.html) and [java.util.logging](https://logging.apache.org/log4j/2.x/log4j-jul/index.html), as far as I know.

joschi:
@ivarconr Should we continue with this pull request or would you rather not do the change and I close it?

ivarconr:
I have been holding off as it probably should be part of v4.

joschi:
@ivarconr Yes, I guess that makes sense since it's a breaking change (with regards to the dependencies).

Do you have an idea when you would like to start working on version 4 of Unleash or the Unleash Java client?

ivarconr:
@joschi I have already started on both. 

Variants and strategy constraints are both v4 functionality we have started on testing out with clients to make sure they work correctly.

jan-berge-ommedal:
@ivarconr @joschi  i am interested in this change as well, and i see an ongoing 4.x branch. Perhaps this PR could re-target and merge into that branch?

joschi:
@jan-berge-ommedal @ivarconr I'm not using Unleash anymore and this PR seems to have deteriorated quite a bit since opening it.

Additionally, the `4.x` branch hasn't seen any activity for 2 years, so I'm probably going to close this PR.

Feels free to use my code, though. 😄

jan-berge-ommedal:
Hi, getting back on this horse again.

@ivarconr what are your thoughts about this one? 

I personally favor slf4j-api over log4j-api since it is lighter and more flexible. But logging in java quickly gets messy anyways, so the ideal library in my mind is the one that comes with no hard (log) dependencies and lets you take control yourself. Optionally with some sensible defaults. 

So i really like this idea https://github.com/Unleash/unleash-client-java/pull/78#issuecomment-503270519

The unleash client could come with no dependency on neither slf4j-api nor log4j-api, try to auto-detect them at runtime and fall back on sysout logging if all else fails. If it in addition provides a way to bridge logging over the subscriber-api, it will be very non-obtrusive and easy to integrate everywhere.

If this sounds great to you @ivarconr, I can make a suggestion that might even be safe to put into v3 since it will auto-detect log4j-api.

ivarconr:
> If this sounds great to you @ivarconr, I can make a suggestion that might even be safe to put into v3 since it will auto-detect log4j-api.

It sounds like a decent solution to me.

ivarconr:
closing this as change-set is rather big. We will create a new PR which only do the sl4j change. Relevant change: https://github.com/Unleash/unleash-client-java/pull/78/commits/d5013c7c8aa113dbceeb47c960e73c0212b2407c

PR #93, opened 2019-11-12
feat: Add support for fallback-action 

dennis-menge:
Today it is possible to define a fallback value which the `isEnabled` will return if the feature toggle is not defined.

We want to also allow the user to send in a fallbackFunction instead of a fallback value (which will take precedence over fallback value).

The fallback function should get to arguments injected: `toggleName` and the enhanced `unleashContext`. The function is implemented by the caller and should return a boolean value. 

Example on how it would work:
`isEnabled("someToggle", context, (toggleName, eContext) -> false)`

### DIFF
--- a/src/main/java/no/finn/unleash/DefaultUnleash.java
+++ b/src/main/java/no/finn/unleash/DefaultUnleash.java
@@ -85,33 +85,21 @@ public final class DefaultUnleash implements Unleash {
 
     @Override
     public boolean isEnabled(final String toggleName, final UnleashContext context, final boolean defaultSetting) {
-        return isEnabled(toggleName, context, defaultSetting, (name, unleashContext) -> defaultSetting);
+        return isEnabled(toggleName, context, (n, c) -> defaultSetting);
     }
 
+    @Override
     public boolean isEnabled(final String toggleName, final BiFunction<String, UnleashContext, Boolean> fallbackAction) {
-        return isEnabled(toggleName, false, fallbackAction);
+        return isEnabled(toggleName, contextProvider.getContext(), fallbackAction);
     }
 
-    public boolean isEnabled(final String toggleName, boolean defaultSetting, final BiFunction<String, UnleashContext, Boolean> fallbackAction) {
-        return isEnabled(toggleName, contextProvider.getContext(), defaultSetting, fallbackAction);
-    }
-
-    public boolean isEnabled(final String toggleName, UnleashContext context, boolean defaultSetting, final BiFunction<String, UnleashContext, Boolean> fallbackAction) {
+    public boolean isEnabled(String toggleName, UnleashContext context, BiFunction<String, UnleashContext, Boolean> fallbackAction) {
         FeatureToggle featureToggle = toggleRepository.getToggle(toggleName);
-
-        return isEnabled(toggleName, featureToggle, context, defaultSetting, fallbackAction);
-    }
-
-    private boolean isEnabled(String toggleName, FeatureToggle featureToggle, UnleashContext context, boolean defaultSetting, BiFunction<String, UnleashContext, Boolean> fallbackAction) {
         boolean enabled;
         UnleashContext enhancedContext = context.applyStaticFields(config);
 
         if (featureToggle == null) {
-            enabled = defaultSetting;
-
-            if(fallbackAction != null) {
-                enabled = fallbackAction.apply(toggleName, enhancedContext);
-            }
+            enabled = fallbackAction.apply(toggleName, enhancedContext);
         } else if(!featureToggle.isEnabled()) {
             enabled = false;
         } else if(featureToggle.getStrategies().size() == 0) {
@@ -133,7 +121,7 @@ public final class DefaultUnleash implements Unleash {
     @Override
     public Variant getVariant(String toggleName, UnleashContext context, Variant defaultValue) {
         FeatureToggle featureToggle = toggleRepository.getToggle(toggleName);
-        boolean enabled = isEnabled(toggleName, context, false);
+        boolean enabled = isEnabled(toggleName, context, (n, c) -> false);
         Variant variant = enabled ? selectVariant(featureToggle, context, defaultValue) : defaultValue;
         metricService.countVariant(toggleName, variant.getName());
         return variant;
--- a/src/main/java/no/finn/unleash/DefaultUnleash.java
+++ b/src/main/java/no/finn/unleash/DefaultUnleash.java
@@ -1,11 +1,5 @@
 package no.finn.unleash;
 
-import java.util.Arrays;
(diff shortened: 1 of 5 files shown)


src/main/java/no/finn/unleash/Unleash.java
```diff
@@ -15,6 +15,10 @@ default boolean isEnabled(String toggleName, UnleashContext context, boolean def
         return isEnabled(toggleName, defaultSetting);
     }
 
+    boolean isEnabled(final String toggleName, boolean defaultSetting, final FallbackAction fallbackAction);
```
ivarconr: Instead of defining a new interface we could actually use one of the functional interfaces: `BiFunction<String, UnleashContext, Boolean>` instead.

src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -73,12 +77,33 @@ public boolean isEnabled(final String toggleName, final boolean defaultSetting)
...
+    }
+
+    public boolean isEnabled(final String toggleName, boolean defaultSetting, final FallbackAction fallbackAction) {
+        FeatureToggle featureToggle = toggleRepository.getToggle(toggleName);
+
+        if(featureToggle == null && fallbackAction != null) {
+            fallbackAction.apply(toggleName, contextProvider.getContext());
+        }
+
+        return isEnabled(featureToggle, contextProvider.getContext(), defaultSetting);
```
ivarconr: I would rather invert it and have this function being the master and have the other `isEnabled` calls send in a proper fallback-action defined.

src/main/java/no/finn/unleash/Unleash.java
dennis-menge: Oh yeah, completely forgot about these, I will change that accordingly.

src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -74,11 +71,28 @@ public boolean isEnabled(final String toggleName, final boolean defaultSetting)
...
+        return isEnabled(toggleName, false, fallbackAction);
+    }
+
+    public boolean isEnabled(final String toggleName, boolean defaultSetting, final BiFunction<String, UnleashContext, Boolean> fallbackAction) {
+        return isEnabled(toggleName, contextProvider.getContext(), defaultSetting, fallbackAction);
+    }
+
+    public boolean isEnabled(final String toggleName, UnleashContext context, boolean defaultSetting, final BiFunction<String, UnleashContext, Boolean> fallbackAction) {
         FeatureToggle featureToggle = toggleRepository.getToggle(toggleName);
         boolean enabled = isEnabled(featureToggle, context, defaultSetting);
```
ivarconr: I believe it would be better if we modify the private `isEnabled`-method (line 99) to take `FallbackAction` as last input and have all public `isEnabled`-methods call this one. There are some reasons for this:

- Avoid duplication of logic. 
- ensure that FallbackAction gets the enhanced context injected. 

This change would require existing `isEnabled`-methods to also send in a fallback function, this can be achieved via a lambda that just returns the fallbackValue injected to the method (or false)

src/main/java/no/finn/unleash/Unleash.java
```diff
@@ -15,6 +16,14 @@ default boolean isEnabled(String toggleName, UnleashContext context, boolean def
         return isEnabled(toggleName, defaultSetting);
     }
 
+    default boolean isEnabled(final String toggleName, final BiFunction<String, UnleashContext, Boolean> fallbackAction) {
+        return isEnabled(toggleName, false, fallbackAction);
+    }
+
+    boolean isEnabled(final String toggleName, boolean defaultSetting, final BiFunction<String, UnleashContext, Boolean> fallbackAction);
```
ivarconr: By not providing fallback for new interface methods we will break existing implementations of the interface. This is probably ok, but needs to be at least a minor.

dennis-menge: Added a default implementation like for the other methods

ivarconr:
This is really closing in to something good now :rocket:. 

I have limited time today, but how to be able to do a proper validation tomorrow!

Feel free to also [join us on slack](https://join.slack.com/t/unleash-community/shared_invite/enQtNjUxMjU2MDc0MTAxLTJjYmViYjkwYmE0ODVlNmY1YjcwZGRmZWU5MTU1YTQ1Nzg5ZWQ2YzBlY2U1MjlmZDg5ZDRmZTMzNmQ5YmEyOGE) for more in depth discussions:

src/main/java/no/finn/unleash/Unleash.java
```diff
@@ -15,6 +16,18 @@ default boolean isEnabled(String toggleName, UnleashContext context, boolean def
         return isEnabled(toggleName, defaultSetting);
     }
 
+    default boolean isEnabled(final String toggleName, final BiFunction<String, UnleashContext, Boolean> fallbackAction) {
+        return isEnabled(toggleName, false, fallbackAction);
+    }
+
+    default boolean isEnabled(String toggleName, UnleashContext context, boolean defaultSetting, BiFunction<String, UnleashContext, Boolean> fallbackAction) {
```
ivarconr: This one is not needed. It does not make sense to allow both defaultSetting and a fallbackAction.

ivarconr:
Hi again @dennis-menge. I did a few adjustment to this PR. Could you look it over and see if iy make sense?

https://github.com/Unleash/unleash-client-java/pull/93/commits/e96a1730d2803ac66e2941e2e67b0b698b438a43

src/main/java/no/finn/unleash/FakeUnleash.java
```diff
@@ -27,6 +28,19 @@ public boolean isEnabled(String toggleName, boolean defaultSetting) {
...
+        return isEnabled(toggleName, fallbackAction);
+    }
+
+    @Override
+    public boolean isEnabled(String toggleName, BiFunction<String, UnleashContext, Boolean> fallbackAction) {
+        if(!features.containsKey(toggleName)) {
+            return fallbackAction.apply(toggleName, UnleashContext.builder().build());
+        }
+        return isEnabled(toggleName);
+    }
```
ivarconr: I believe this is all that is needed for the fake-unleash.

dennis-menge:
> Hi again @dennis-menge. I did a few adjustment to this PR. Could you look it over and see if iy make sense?
> 
> [e96a173](https://github.com/Unleash/unleash-client-java/commit/e96a1730d2803ac66e2941e2e67b0b698b438a43)

Hi Ivar, looks nice, I misunderstood all whole time that defaultSetting and fallbackAction should be provided at the same time with fallbackAction taking precedence, which obviously makes no sense.

Thank you!

PR #96, opened 2019-12-04
Fix java.lang.NullPointerException when no variants defined

blankg:
java.lang.NullPointerException: null	
at no.finn.unleash.variant.VariantUtil.selectVariant(VariantUtil.java:52) ~[unleash-client-java-3.2.10.jar:3.2.10]

When there are feature toggles with no variants defined if you try to do getVariant from DefaultUnleash. this is because when you do deserializeVersion1 variants is null and not un empty collection.


ivarconr:
Wow, strange nobody have experienced this problem before.

Thanks!

PR #103, opened 2020-03-12
fix: add `shutdown()` to `Unleash`

Enrico2:
This PR introduces a new `shutdown()` method to 2 interfaces, `Unleash` and `UnleashScheduledExecutor`

In order to properly shutdown a server that's running an Unleash client, all scheduled threads should be shutdown gracefully.

With the current implementation there is no clean way to shutdown the `UnleashScheduledExecutor`. If we're using the default implementation of `UnleashScheduledExecutorImpl`, all fields that would allow us to do so are either private, or are swallowed; The `ScheduledFuture` object returned from `setInterval` is not set to any field in the 2 constructors that call it, `UnleashMetricServiceImpl` and `FeatureToggleRepository`. Nor is the `ScheduledThreadPoolExecutor` exposed (it has a `shutdown()` method).

Our only option right now is to implement our own `UnleashScheduledExecutor`, pass it in as the config, and have a `close()` or `shutdown()` method on it. (we can't even use `UnleashScheduledExecutorImpl` because of the private fields.

The implementation gets the executor from the `config` instance and call `shutdown()` on it. That way, no matter which implementation of `UnleashScheduledExecutor` one would use, they can shut it down via the main interface.

### DIFF
--- a/src/main/java/no/finn/unleash/DefaultUnleash.java
+++ b/src/main/java/no/finn/unleash/DefaultUnleash.java
@@ -162,4 +162,9 @@ public final class DefaultUnleash implements Unleash {
     private Strategy getStrategy(String strategy) {
         return strategyMap.containsKey(strategy) ? strategyMap.get(strategy) : UNKNOWN_STRATEGY;
     }
+
+    @Override
+    public void shutdown() {
+        config.getScheduledExecutor().shutdown();
+    }
 }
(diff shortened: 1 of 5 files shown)


Enrico2:
@ivarconr thanks! I'll keep an eye out to see when you publish an new version ;)

Issue #94, opened 2019-11-12
feat: function for returning headers


ivarconr:
This would allow for dynamic header setting, which is important to enable rotation of auth-tokens etc without restarting. 

For reference, this feature is already implemented in the Node.js SDK, see https://github.com/Unleash/unleash-client-node/issues/151

My thought was that it should be possible to register a function as part of the Unleash Configuration which would return a Map (name, value) of headers to add request sent to the unleash-server. If this function is set it should take precedence over `customHttpHeaders` already part of the UnleashConfig.java.

PR #104, opened 2020-03-13
fix: add a custom http headers provider for unleash config class

ajvasey:
This will make it possible to acquire a new or refresh auth tokens when talking to the server.  This also addresses the following issue.  https://github.com/Unleash/unleash-client-java/issues/94


ivarconr:
Thanks! Looks clean and nice.

ivarconr:
Thanks for your contribution @ajvasey 

Released as part of 3.3.2

PR #115, opened 2020-12-08
fix: follow redirect once

### LINKED ISSUE #113: Exception at HTTP redirect
Some JAVA Clients use the unleash endpoint with **HTTP**: `http://toggle.example.org/api`. Today we did migrate the feature-toggle server and enforced a SSL redirect (308) to the very same HTTPS endpoint. The JAVA client throws an exception:
```
no.finn.unleash.UnleashException · Error fetching toggles from Unleash API - StatusCode: 308, Location: https://toggle.example.org/api/client/features
```
i think this is the code related to this:
https://github.com/Unleash/unleash-client-java/blob/12ed9c286264437efb752245c28f11c0eb01f2f2/src/main/java/no/finn/unleash/repository/HttpToggleFetcher.java#L45-L52

I expected that the unleash client would follow the redirects of `307` or `308` redirectsand fetch the toggles from there.

michbeck100:
This PR adds the ability to follow redirects (HTTP 301 - 303). To prevent infinite loops this is done only once. This PR fixes #113

### DIFF
--- a/src/main/java/no/finn/unleash/repository/HttpToggleFetcher.java
+++ b/src/main/java/no/finn/unleash/repository/HttpToggleFetcher.java
@@ -28,28 +28,10 @@ public final class HttpToggleFetcher implements ToggleFetcher {
     public FeatureToggleResponse fetchToggles() throws UnleashException {
         HttpURLConnection connection = null;
         try {
-            connection = (HttpURLConnection) toggleUrl.openConnection();
-            connection.setConnectTimeout(CONNECT_TIMEOUT);
-            connection.setReadTimeout(CONNECT_TIMEOUT);
-            connection.setRequestProperty("Accept", "application/json");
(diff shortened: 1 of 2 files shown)


ivarconr:
Thanks! This looks good! ;)

ivarconr:
One thing; 
I think it is good to follow a redirect once, but; if the the server always return a 301 redirect for every fetch, would it then make sense to update the local toggle URL? If not every fetch will end up sending two requests, which sounds a bit stupid. 

I'm not convinced the extra complexity is worth it. It could also be useful to just log the situation as a warning. 

What do you think @michbeck100?

michbeck100:
I would also say that logging should be enough. In the end, the local toggle URL should be changed by the user, because it originates from the config. I've updated the PR accordingly.

PR #117, opened 2021-01-28
feat: Add support for filtering toggles on project

chriswk:
This will have no effect until the server is released with support for this, but it's good to be ahead of time :)

### DIFF
--- a/src/main/java/no/finn/unleash/util/UnleashConfig.java
+++ b/src/main/java/no/finn/unleash/util/UnleashConfig.java
@@ -26,6 +26,7 @@ public class UnleashConfig {
     private final String instanceId;
     private final String sdkVersion;
     private final String backupFile;
+    private final String projectName;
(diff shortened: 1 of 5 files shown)


src/main/java/no/finn/unleash/util/UnleashConfig.java
```diff
@@ -26,6 +26,7 @@
     private final String instanceId;
     private final String sdkVersion;
     private final String backupFile;
+    private final String projectName;
```
ivarconr, line 29: So we have no intention of a SDK to be able to be part of "multiple projects"?

chriswk: After discussion offline, for now this is kept as a String, not a collection of Strings. However, we've made the UnleashConfig constructor private, so we can change this without making another breaking change if we see the need for adding support for multiple projects.

Issue #86, opened 2019-08-20
ClientRegistration Event - Suggestions


Green-Apple:
Small change suggestion. 
You may want to change getInterval() to getSendMetricInterval() as this could easily be confused with the fetchTogglesInterval value from the UnleashConfig if the source code isn't examined.

Also, I noticed that the ClientRegistration is pulling a several values from the UnleashConfig.  Would it make more sense to instead just store the entire UnleashConfig as an event variable instead of the few selected fields? That way, someone implementing the clientRegistered() method of the UnleashSubscriber can pick/choose which fields are relevant for them to log.

ivarconr:
I agree with your suggestions. Will look in to it soon, when I have some time for it. 

Feel free to contribute a PR for it, but be careful about breaking changes. (E.g. we probably should deprecate "gerInterval' when adding the new method).

chriswk:
After trying to write a PR and thinking about this a bit more, I think this is a wontfix.

UnleashConfig has a scheduled executor and a reference to a subscriber that doesn't lend itself easily to serialization. As such it doesn't make sense to include the entire config object in the event.

If there are specific fields from the config you're missing, feel free to submit a PR adding those fields to the event you need it in. But as of now, we won't include the entire config in the event.

chriswk:
With regards to interval vs sendMetricsInterval, we do agree that the name may be confusing. However, the client specification requires the field to be serialized as interval for now. We might revisit that for a later major release.

PR #120, opened 2021-01-28
Add UnleashConfig field to ClientRegistration events

chriswk:
I'm really unsure about the fields on the class, I think it still makes sense to keep the fields we get straight from the config, because that adds value to the serialized version of the class (Gson serializes the fields to a json object)

I needed to add an exclusion strategy to prevent serialization of the config field because UnleashConfig apparently has some recursive fields, because it blew my stack away while running the client specification tests

fixes: #86

### DIFF
--- a/src/main/java/no/finn/unleash/metric/ClientRegistration.java
+++ b/src/main/java/no/finn/unleash/metric/ClientRegistration.java
@@ -16,6 +16,7 @@ public class ClientRegistration implements UnleashEvent {
     private final long interval;
     private final String environment;
 
+
     ClientRegistration(UnleashConfig config, LocalDateTime started, Set<String> strategies) {
         this.environment = config.getEnvironment();
         this.appName = config.getAppName();
@@ -46,7 +47,7 @@ public class ClientRegistration implements UnleashEvent {
         return started;
     }
 
-    public long getInterval() {
+    public long getSendMetricsInterval() {
         return interval;
     }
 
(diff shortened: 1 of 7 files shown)


src/main/java/no/finn/unleash/metric/ClientRegistration.java
```diff
@@ -44,8 +50,8 @@ public LocalDateTime getStarted() {
         return started;
     }
 
-    public long getInterval() {
```
ivarconr, line 47: I needs to serialize as "interval" as this is the contract on the API side:
https://github.com/Unleash/unleash/blob/master/lib/services/client-metrics/register-schema.js#L17

chriswk: ah, good point.

ivarconr:
Yes, this solves the need, but I'm not too fan of having excluded json-fields on the object.

chriswk:
Agreed, I didn't like it too much, which I guess might explain why we've extracted the fields earlier.

src/main/java/no/finn/unleash/metric/ClientRegistration.java
chriswk: It should now be serialized as interval still.

chriswk:
Should we drop including it?

ivarconr:
> Should we drop including it?

I was thinking if we could solve this in a slightly different way? 

Possible options:
1. Have a special config event
2. Include the config as part of the event itself when publishing?

chriswk:
I'll have a look tomorrow.

chriswk:
I'm tending towards creating a new config event. If we start requiring UnleashConfig as a separate argument to the publish methods, we'd also require all events to store the config as a field. This would make us need to use the JsonExclude annotation in all events, defeating the purpose of looking at this again.

ivarconr:
> I'm tending towards creating a new config event.

I agree.

chriswk:
See comment in #86 - We won't do this for now.

PR #121, opened 2021-01-28
fix: Add environment to metrics

### LINKED ISSUE #119: fix: add environment to metrics
https://github.com/Unleash/unleash-client-java/blob/master/src/main/java/no/finn/unleash/metric/ClientMetrics.java

ivarconr:
closes #119


chriswk (approved):
LGTM - seems like a nice addition allowing server side to display from what environment we see client activity.

ivarconr:
> LGTM - seems like a nice addition allowing server side to display from what environment we see client activity.

Yes, we still need to collect this information on the server-side to make it useful. Will create a small PR for that as well

Issue #114, opened 2020-09-16
Kotlin support for Java client


alxsimo:
How about adding Kotlin support for the Java client?

I already saw there is a Kotlin migrated client, it's just migrated, no extra features added.
The power of Kotlin comes from extension functions, and we could have inside this project, a separated source set for Kotlin extensions and maybe even migrate some code.

ivarconr:
This is an interesting idea. Kotlin has a few interesting capabilities that could improve create even better and more advanced API surfaces for the users. 

I think it still is good to be able to ship pure java SDK without Kotlin specific dependencies.

chriswk:
To make it better to use the java client from kotlin, we could consider adding @Nullable and @NotNullable annotations to our method parameters, allowing Kotlin users to make the compiler do null-checks for them.

ivarconr:
Would it make sense to have a separate Kotlin SDK repo? I assume that we would quite different API surface. We could possible share some of the basic building blocks, such as the strategy implementation.

checketts:
No, there is no need to create a separate repo. Any Kotlin specific pieces can co-exist without causing trouble for the Java users. I've created similar Kotlin extensions for other projects and would consider submitting a PR if this sounds reasonable. The key is marking the kotlin sdk as an 'optional' Maven dependency (Since Kotlin projects will already have it).

I would do a first pass adding the `@Nullable` annotations then do a subsequent PR that adds a Kotlin DSL.

checketts:
Current Kotlin DSL that I've been playing with:

##Make the UnleashConfig builder use a receiver block so you can call builder methods in that block and make the required parameters part of the base call (named parameter work there as well)
```
 val config = unleashConfig("bob","http://localhost:4242/api/" ) {
            unleashContextProvider(contextProvider)
        }
```        

Feature enabled check options:
```
//Normal 'if' block, no changes
        if(unleash.isEnabled("test")) {
            print("yes")
        } else {
            print("no")
        }

//Create an extra 'infix' notation, I don't love it, but if there is a good infix notation that could make feature checks stand out more I think this could be good
        if(unleash feature "test") {
            print("yes")
        } else {
            print("no")
        }

//When notation also doesn't require any DSL changes. I don't find it much more succinct than the simple 'if'
        when(unleash.isEnabled("test")) {
            true -> print("yes")
            else -> print("no")
        }
```

checketts:
Here is on Kotlin-ism that I've begun using.

Given the extension function:

[...]

PR #122, opened 2021-02-04
Add Nullable annotations where applicable

checketts:
In support for #114 

To avoid adding `@Nonnull` annotations everywhere, this PR adds `package-info.java` files and essentially marks the entire package as `@Nonnull` for return types, parameters, and fields. The remaining changes then only have to call out where nullability is allowed by adding the `@Nullable` annotation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,6 +67,13 @@
         </dependency>
 
         <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+            <version>3.0.2</version>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
             <groupId>com.sangupta</groupId>
             <artifactId>murmur</artifactId>
             <version>1.0.0</version>
--- a/src/main/java/no/finn/unleash/repository/JsonToggleParser.java
+++ b/src/main/java/no/finn/unleash/repository/JsonToggleParser.java
@@ -20,7 +20,7 @@ final class JsonToggleParser {
                                 ToggleCollection.class, new JsonToggleCollectionDeserializer())
                         .create();
         ToggleCollection gsonCollection = gson.fromJson(reader, ToggleCollection.class);
-        if (gsonCollection == null || gsonCollection.getFeatures() == null) {
+        if (gsonCollection == null) {
             throw new IllegalStateException("Could not extract toggles from json");
         }
         return new ToggleCollection(gsonCollection.getFeatures());
--- a/src/main/java/no/finn/unleash/util/IpAddressMatcher.java
+++ b/src/main/java/no/finn/unleash/util/IpAddressMatcher.java
@@ -15,6 +15,8 @@
  */
 package no.finn.unleash.util;
 
+import no.finn.unleash.lang.Nullable;
+
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.Arrays;
@@ -31,7 +33,7 @@ import java.util.regex.Pattern;
 public final class IpAddressMatcher {
     private static final Pattern SPLITTER = Pattern.compile("/");
     private final int nMaskBits;
-    private final InetAddress requiredAddress;
+    @Nullable private final InetAddress requiredAddress;
 
     /**
      * Takes a specific IP address or a range specified using the IP/Netmask (e.g. 192.168.1.0/24 or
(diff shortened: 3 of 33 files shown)


pom.xml
```diff
@@ -66,6 +66,13 @@
             <version>${version.slf4j}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+            <version>3.0.2</version>
+            <optional>true</optional>
```
checketts, line 73: Note this is an optional dependency addition so it won't impact downstream consumers.

src/main/java/no/finn/unleash/repository/JsonToggleParser.java
```diff
@@ -20,7 +20,7 @@ public static ToggleCollection fromJson(Reader reader) throws IllegalStateExcept
                                 ToggleCollection.class, new JsonToggleCollectionDeserializer())
                         .create();
         ToggleCollection gsonCollection = gson.fromJson(reader, ToggleCollection.class);
-        if (gsonCollection == null || gsonCollection.getFeatures() == null) {
```
checketts, line 23: The annotations (and the code) demonstated the second null check wasn't needed.

src/main/java/no/finn/unleash/util/IpAddressMatcher.java
```diff
@@ -52,14 +54,14 @@ public IpAddressMatcher(String ipAddress) {
...
-    public boolean matches(String address) {
+    public boolean matches(@Nullable String address) {
         if (address == null || address.isEmpty() || requiredAddress == null) {
             return false;
         }
 
         InetAddress remoteAddress = parseAddress(address);
 
-        if (!requiredAddress.getClass().equals(remoteAddress.getClass())) {
+        if (remoteAddress == null || !requiredAddress.getClass().equals(remoteAddress.getClass())) {
```
checketts, line 64: Annotations uncovered that `parseAddress` returns a `null` in some cases. Added the check to account for that. Please check the logic is correct.

src/main/java/no/finn/unleash/util/UnleashConfig.java
```diff
@@ -36,23 +37,23 @@
...
             CustomHttpHeadersProvider customHttpHeadersProvider,
-            String appName,
+            @Nullable String appName,
             String environment,
-            String instanceId,
+            @Nullable String instanceId,
             String sdkVersion,
             String backupFile,
-            String projectName,
+            @Nullable String projectName,
```
checketts, line 48: This constructor has checks for all nullable parameters except the `projectName`. I assume it is OK to be nullable (based on current behavior), but wanted to point out the difference.

src/main/java/no/finn/unleash/variant/VariantUtil.java
```diff
@@ -61,7 +62,10 @@ private static String getIdentifier(UnleashContext context) {
     }
 
     public static Variant selectVariant(
-            FeatureToggle featureToggle, UnleashContext context, Variant defaultVariant) {
+        @Nullable FeatureToggle featureToggle, UnleashContext context, Variant defaultVariant) {
+        if(featureToggle == null) {
```
checketts, line 66: Annotations revealed that the `featureToggle` could be passed in as a null, so adding this check to account for that. Please check that returning the `defaultVariant` is the correct response.

src/main/java/no/finn/unleash/util/UnleashConfig.java
chriswk: Good point, yes, it's ok to be nullable, since only enterprise has support for projects at the moment

pom.xml
chriswk: Fair enough :)

src/main/java/no/finn/unleash/util/UnleashConfig.java
chriswk: Also because there are situations where you might not want to filter on projectName, leaving it out is perfectly fine.

src/main/java/no/finn/unleash/variant/VariantUtil.java
chriswk: Looking at the places we call this, I'd say this should be the correct response

chriswk (approved):
Thanks so much for this. Good job, you just saved me some work I wasn't particularly looking forward to :)

chriswk:
@checketts  - Could you add a quick changelog line to CHANGELOG.md pointing out that nullable/nonnullable annotations are added?

checketts:
Changelog updates added

chriswk:
Excellent. Thank you :)

PR #123, opened 2021-02-06
fix: Add logging for when a matching strategy is not found

checketts:
I created a new custom activation in Unleash and was very confused as to why it wasn't taking place. People familiar with Unleash would have know what was wrong, in that I hadn't created/registered a corresponding strategy.

This PR adds a warning so it is clear in the calling app which toggle/strategy is affected and how they can address it.

### DIFF
--- a/src/main/java/no/finn/unleash/DefaultUnleash.java
+++ b/src/main/java/no/finn/unleash/DefaultUnleash.java
@@ -21,8 +21,11 @@ import no.finn.unleash.repository.ToggleBackupHandlerFile;
 import no.finn.unleash.repository.ToggleRepository;
 import no.finn.unleash.strategy.*;
 import no.finn.unleash.util.UnleashConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class DefaultUnleash implements Unleash {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultUnleash.class);
     private static final List<Strategy> BUILTIN_STRATEGIES =
             Arrays.asList(
                     new DefaultStrategy(),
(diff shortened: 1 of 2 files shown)


ivarconr (approved):

checketts:
I had forgotten to add a changelog entry. That is addressed now

src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -120,12 +123,20 @@ private boolean checkEnabled(
...
-                                    as ->
-                                            getStrategy(as.getName())
-                                                    .isEnabled(
-                                                            as.getParameters(),
-                                                            enhancedContext,
-                                                            as.getConstraints()));
+                                    strategy -> {
+                                        Strategy configuredStrategy = getStrategy(strategy.getName());
+                                        if (configuredStrategy == UNKNOWN_STRATEGY) {
+                                            LOGGER.warn(
```
ivarconr, line 129: I think this is fine, it is good to inform the user/developer that there is something missing, but would just like to get someones second opinion: 

Is it ok to log this error every single time "isEnabled" is called? Or should we try to only log it once? I am quite not sure what the right answer is, but I can easily see the application log being blown up by a simple miss-configuration in the UI.

checketts: I think logging every time is preferred to display why something isn't working. Especially to get in this state, it isn't a simple typo in the Admin UI (you have to create a new strategy, then assign it as an activation to a toggle)

If a team does this commonly (I can't imagine why) They could mute the logger completely.

ivarconr: thanks. this is a valid point.

Issue #124, opened 2021-02-07
Support for precalcuating toggles and manual metrics


checketts:
I have a number of SPA (Single Page Applications) that will run in a user's browser. The code in the browser is written in Typescript (Angular or React)

I would like to add 2 capabilities: 

# Precalculated toggles
A way to pre-calculate all toggles for a given UnleashContext without triggering the corresponding metrics

This will let me generate the results of toggles to send to the browser which can then be used by the client side code.

# Manually sending metrics

Periodically when the front-end checks back for toggle updates it can report its metrics to the backend which will be able to forward them on to Unlease.

I don't want my frontend code calling Unleash directly due to differing authentication needs and scalability.

Would this capability be welcome as a PR?

ivarconr:
Hi @checketts,

As part of the commercial offering, we already have the [unleash-proxy](https://www.unleash-hosted.com/articles/the-unleash-proxy), providing this functionalities (and more) on top of the [Node SDK](https://github.com/Unleash/unleash-client-node). 

Our plan is to also open-source the Unleash Proxy for everyone, but that would take a few more sprints before we are able to prioritize this work (it will require us to do some cleanup to make it 100% compatible with the community version of Unleash). You might want to use the Unleash Proxy as the official way to integrate with SPAs

That being said, I do not see any harm of adding these capabilities for the Java SDK as well.

checketts:
Thanks for the tip about unleash-proxy! I'll keep an eye out for when it becomes open source. 

I've made a minimal PR to the Java client that I believe will allow me to move forward for now.

checketts:
Closing since this is addressed in version 4.1.0

PR #125, opened 2021-02-08
feat: Add ability to disable metrics via UnleashContext

checketts:
Addresses #124 

Refined approach with `more` concept for evaluating all toggles as well as counting metrics.

Example usage:
```
     @PostMapping("/features")
    @ResponseBody
    public List<EvaluatedToggle> featureMetrics(@RequestBody List<FeatureUsage> usage){
        usage.forEach(toggle -> {
            for (int i = 0; i < toggle.count; i++) {
                unleash.more().count(toggle.getName(), toggle.isEnabled());
            }
        });

        UnleashContext cont = createContext();
        return unleash.more().evaluateAllToggles(cont);
    }
```

This returns a json of features, I might refine this response in my service to strip out nulls/empty variants:
```
[
  {
    "enabled": true,
    "name": "feature",
    "variant": {
      "name": "disabled",
      "payload": {
        "type": "string",
        "value": null
      },
      "enabled": false
    }
  },
  {
    "enabled": false,
    "name": "other-feature",
    "variant": {
      "name": "disabled",
      "payload": {
        "type": "string",
        "value": null
      },
      "enabled": false
    }
  },
  {
    "enabled": false,
    "name": "user-feature",
    "variant": {
      "name": "disabled",
      "payload": {
        "type": "string",
        "value": null
      },
      "enabled": false
    }
  }
]
```

And receives a list of feature checks so we can poke the metrics:
```
[{
"name": "myToggle",
"enabled": true,
"count": 15
}]
```

(I apologize for the code formatting commit, but maven:spotless insisted they were needed, so I included them. I can separate them out if you want)

### DIFF
--- a/src/main/java/no/finn/unleash/DefaultUnleash.java
+++ b/src/main/java/no/finn/unleash/DefaultUnleash.java
@@ -10,6 +10,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
 import no.finn.unleash.event.EventDispatcher;
 import no.finn.unleash.event.ToggleEvaluated;
 import no.finn.unleash.lang.Nullable;
@@ -58,13 +60,29 @@ public final class DefaultUnleash implements Unleash {
             UnleashConfig unleashConfig,
             ToggleRepository toggleRepository,
             Strategy... strategies) {
+        this(unleashConfig,
+            toggleRepository,
+            buildStrategyMap(strategies),
+            unleashConfig.getContextProvider(),
+            new EventDispatcher(unleashConfig),
+            new UnleashMetricServiceImpl(unleashConfig, unleashConfig.getScheduledExecutor())
+            );
+    }
+
+    // Visible for testing
+    public DefaultUnleash(
+        UnleashConfig unleashConfig,
+        ToggleRepository toggleRepository,
+        Map<String, Strategy> strategyMap,
+        UnleashContextProvider contextProvider,
+        EventDispatcher eventDispatcher,
+        UnleashMetricService metricService) {
         this.config = unleashConfig;
         this.toggleRepository = toggleRepository;
-        this.strategyMap = buildStrategyMap(strategies);
-        this.contextProvider = unleashConfig.getContextProvider();
-        this.eventDispatcher = new EventDispatcher(unleashConfig);
-        this.metricService =
-                new UnleashMetricServiceImpl(unleashConfig, unleashConfig.getScheduledExecutor());
+        this.strategyMap = strategyMap;
+        this.contextProvider = contextProvider;
+        this.eventDispatcher = eventDispatcher;
+        this.metricService = metricService;
         metricService.register(strategyMap.keySet());
     }
 
@@ -159,15 +177,24 @@ public final class DefaultUnleash implements Unleash {
         return ofNullable(toggleRepository.getToggle(toggleName));
     }
 
+    /**
+     * Use more().getFeatureToggleNames() instead
+     * @return a list of known toggle names
+     */
+    @Deprecated()
     public List<String> getFeatureToggleNames() {
         return toggleRepository.getFeatureNames();
     }
 
+    /**
+     * Use more().count() instead
+     */
+    @Deprecated
     public void count(final String toggleName, boolean enabled) {
         metricService.count(toggleName, enabled);
     }
 
-    private Map<String, Strategy> buildStrategyMap(@Nullable Strategy[] strategies) {
+    private static Map<String, Strategy> buildStrategyMap(@Nullable Strategy[] strategies) {
         Map<String, Strategy> map = new HashMap<>();
 
         BUILTIN_STRATEGIES.forEach(strategy -> map.put(strategy.getName(), strategy));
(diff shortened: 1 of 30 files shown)


chriswk (approved):
LGTM, and I see you and Ivar already hashed out some details in #124.

Bit surprised why we would need more spotless after my way too big formatting commit, but I guess that's the way when you allow automatic tools to do the job, they aren't always perfect.
No problem for me to include it in the PR, it's still clear to me what the actual change is.

Appreciate the PR, @ivarconr will also have a gander before we merge it. :)

checketts:
The spotless updates were fallout from my Nullability PR earlier. So they were legitimate changes.

I had forgotten to add a changelog entry. That is addressed now. However once #123 is merged it will probably create a merge-conflict, so I'll be sure to rebase to resolve that when it happens.

ivarconr:
Hi again @checketts,

Really happy you take imitative and sending this PR! 

You have found a very pragmatic solution which require a minimal set of changes to satisfy the need of disabling metrics. I have been thinking about this change a few times today and I have concluded that I do not think this PR represent the correct approach to this problem. I am afraid that the `UnleashContext` object ends up being overloaded with overrides for how the SDK should work. The intended purpose of the `UnleashContext` object is to carry data from the application to the SDK, in order for the SDK to perform the desired action and return an result. 

I would rather see a specific function for this use-case where you send in the UnleashContext as is, and get back a list of evaluated feature toggles for that specific context. I think we can take some inspiration on how the [Unleash-Proxy actually work](https://www.unleash-hosted.com/docs/working-with-the-unleash-proxy/):

How it could look like from code using the SDK:
(I have yet not figured out what to call the `SomeResultObject`)

```java
List<SomeResultObject> result = unleash.evaluateAllToggles(unleashContext);
```

The shape of the List of objects should probably be somehting in the line of:
```

[{
		"enabled": true,
		"name": "SimpleToggle",
		"variant": {
			"enabled": false,
			"name": "disabled"
		}
	},
	{
		"enabled": true,
		"name": "ToggleWithVariant",
		"variant": {
			"enabled": true,
			"name": "brown"
		}
	},
	{
		"enabled": true,
		"name": "ToggleWithVariantAndPayload",
		"variant": {
			"enabled": true,
			"name": "default",
			"payload": {
				"type": "string",
				"value": "Some custom value here!"
			}
		}
	}
]

**Comment by checketts** [comment]
@ivarconr Thanks for the feedback. I'm happy to change approaches and improve the implementation.

To summarize adding directly to the `Unleash` interface is the preferred approach in this case? Or would you prefer to keep that interface as simple as possible and instead add it behind a `batch` operations api?

For example: 
`unleash.evaluateAllToggles()` versus `unleash.batch().evaluateAllToggles()`

The interface returned by the `batch` call would keep the unusual calls from cluttering the common API usecase. (In another project they used `more()` as the name to the advanced sub-api).

**Comment by ivarconr** [comment]
> Or would you prefer to keep that interface as simple as possible and instead add it behind a batch operations api?

This sound very appealing. I have not seen this before, but it feels like the right approach, and possible the `getFeatureToggleNames` fits better behind the batch API as well. 

Would you have any link/examples on the `more()` approach?

**Comment by checketts** [comment]
Micrometer is a metrics facade (similar to slf4j, but for metrics). It has a `more()` concept.

Essentially there are many common ways to use metrics and the advanced or unusual ones that might confuse folks exist in the `more` API.

The method exposing it:
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L454

Implementation:
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L846

I'll take a stab at that approach later today and update the PR along with examples.

**Comment by checketts** [comment]
I updated the original description with usage examples.

File: src/main/java/no/finn/unleash/DefaultUnleash.java [hunk 1]
```diff
@@ -159,15 +177,24 @@ public Variant getVariant(String toggleName, Variant defaultValue) {
...
+    /**
+     * Use more().count() instead
+     */
+    @Deprecated
     public void count(final String toggleName, boolean enabled) {
         metricService.count(toggleName, enabled);
     }
 
-    private Map<String, Strategy> buildStrategyMap(@Nullable Strategy[] strategies) {
+    private static Map<String, Strategy> buildStrategyMap(@Nullable Strategy[] strategies) {
```
ivarconr, line 197: Our of curiosity: what's the benefit of converting this into a private static method?

ivarconr (approved):

ivarconr:
I must say the updated changes looks really good, and it makes perfect sense to start introducing these more advanced use cases as part of a `MoreOptions` interface. Thanks for taking the time to improve the initial implementation. 

I feel comfortable merging and releasing this.

src/main/java/no/finn/unleash/DefaultUnleash.java
checketts: When I added the new constructor for testing, and had the existing constructor forward to the new constructor, the compiler disallowed calling a member function (`buildStrategyMap`) until the object has been constructed. Making the method `static` worked around that, and was harmless since the method doesn't use any member variables anyways.

I noticed that `defaultToggleRepository` on line 48 used the same approach, so I figured it was due to the same reasoning.

chriswk (approved):
Agree with Ivar. This looks like an improvement over the initial attempt and avoids polluting the context with things that aren't actually needed to decide whether or not a toggle is active. I'm also comfortable merging this

PR #127, opened 2021-03-08
feat: Add support for fallback strategy

checketts:
That allows registering a fallback strategy which will support a broader idea of a 'context matching strategy'. I'll create a separate PR for that strategy to keep the PRs focused on each change.

### DIFF
--- a/src/main/java/no/finn/unleash/DefaultUnleash.java
+++ b/src/main/java/no/finn/unleash/DefaultUnleash.java
@@ -219,7 +219,8 @@ public final class DefaultUnleash implements Unleash {
     }
 
     private Strategy getStrategy(String strategy) {
-        return strategyMap.containsKey(strategy) ? strategyMap.get(strategy) : UNKNOWN_STRATEGY;
+        Strategy foundStrategy = strategyMap.getOrDefault(strategy, config.getFallbackStrategy());
+        return foundStrategy != null ? foundStrategy : UNKNOWN_STRATEGY;
     }
 
     @Override
--- a/src/main/java/no/finn/unleash/DefaultUnleash.java
+++ b/src/main/java/no/finn/unleash/DefaultUnleash.java
@@ -38,7 +38,7 @@ public final class DefaultUnleash implements Unleash {
                     new UserWithIdStrategy(),
                     new FlexibleRolloutStrategy());
 
-    private static final UnknownStrategy UNKNOWN_STRATEGY = new UnknownStrategy();
+    public static final UnknownStrategy UNKNOWN_STRATEGY = new UnknownStrategy();
 
     private final UnleashMetricService metricService;
     private final ToggleRepository toggleRepository;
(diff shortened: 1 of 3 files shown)


src/main/java/no/finn/unleash/DefaultUnleash.java
```diff
@@ -219,7 +219,8 @@ public void count(final String toggleName, boolean enabled) {
     }
 
     private Strategy getStrategy(String strategy) {
-        return strategyMap.containsKey(strategy) ? strategyMap.get(strategy) : UNKNOWN_STRATEGY;
+        Strategy foundStrategy = strategyMap.getOrDefault(strategy, config.getFallbackStrategy());
```
ivarconr, line 222: Would it not be better to say that UNKNOWN_STRATEGY is the default fallback strategy, if not overridden?

checketts: Yes. I'll change to that approach. I like that better since it means there will always be a fallback, but I held off changing too much.

checketts:
OK. UNKNOWN_STRATEGY is now the default fallback strategy. Note I made the constant public to allow for that. I could restrict the visibility, but by making it public feels correct in this case since it allows a user to 'reset' it back as a strategy.

src/main/java/no/finn/unleash/util/UnleashConfig.java
```diff
@@ -232,6 +248,7 @@ public boolean isProxyAuthenticationByJvmProperties() {
         private @Nullable UnleashScheduledExecutor scheduledExecutor;
         private @Nullable UnleashSubscriber unleashSubscriber;
         private boolean isProxyAuthenticationByJvmProperties;
+        private Strategy fallbackStrategy;
```
ivarconr, line 251: If we initialize it to to `UNKNOWN_STRATEGY` we can avoid the null checks on line 83:86.

src/main/java/no/finn/unleash/util/UnleashConfig.java
```diff
@@ -333,6 +350,11 @@ public Builder subscriber(UnleashSubscriber unleashSubscriber) {
             return this;
         }
 
+        public Builder fallbackStrategy(@Nullable Strategy fallbackStrategy) {
```
ivarconr, line 353: Generally it seems a bit strange to accept a "null" strategy here. It is not totally clear to me the purpose of that, and is it clear to the users that you then will get the "UNKNOWN" strategy?

src/main/java/no/finn/unleash/util/UnleashConfig.java
checketts: We need a null check at one point to endure the system can't be configured in a bad state. So I can move the null check to the builder setter. 

It pretty common to use null as unsetting a value and saves the user from needing to know what the unknown option is. However if you would prefer I skip that check, I can remove it.

ivarconr: tbh, I do not think it matters much. lets just keep it.

src/main/java/no/finn/unleash/util/UnleashConfig.java
checketts: I don't have a good use case for null here except as error handling/recovery

PR #128, opened 2021-03-08
feat: Add context matching strategy

checketts:
This will be leveraged by #127 as a fallback strategy.  By iterating through the supplied context and matching all that are supplied against the ActivationStrategy parameters, the application will be able to handle new strategies as they are defined, without relying on deploying those new strategies in the codebase.

Example:

If the application is creating a context that includes its appName, userId, environment, userPreferredLanguage, and userRoles, the following strategy could be created in Unleash with the client responding appropriately.

env matching
```
environment = dev,test
```

enabled for admin in lower lanes
```
environment = dev,test
role = admin
```

enabled for English and Spanish speakers
```
userPreferredLanguage = en,es
```

I realize this probably bypasses any need for I would have for 'constraints' that the enterprise offering includes. But since I'm using the open source Unleash this change makes sense. That being said if the Unleash project isn't comfortable including this I would understand (provided #127 support is merged in) Because then I could just include this strategy in each project.

Note this PR is in a draft state because I'm pondering how I want to add percent/rollout to make it clearer, but wanted to expose my thinking to help support #127 discussion.

### DIFF
--- a/src/main/java/no/finn/unleash/strategy/GradualContextMatchingStrategy.java
+++ b/src/main/java/no/finn/unleash/strategy/GradualContextMatchingStrategy.java
@@ -1,0 +1,74 @@
+package no.finn.unleash.strategy;
+
+import static java.util.Arrays.asList;
+import static no.finn.unleash.strategy.GradualRolloutUserIdStrategy.GROUP_ID;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+import no.finn.unleash.UnleashContext;
+import no.finn.unleash.lang.Nullable;
+
+public class GradualContextMatchingStrategy implements Strategy {
+    protected static final String PERCENTAGE = "percentage";
+
+    @Override
+    public String getName() {
+        return "gradualContextMatching";
+    }
+
+    @Override
+    public boolean isEnabled(Map<String, String> parameters) {
+        return false;
+    }
+
+    @Override
+    public boolean isEnabled(Map<String, String> parameters, UnleashContext unleashContext) {
+        String percentageParam = parameters.getOrDefault(PERCENTAGE, parameters.get("rollout"));
+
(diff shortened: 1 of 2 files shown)


checketts:
OK this is ready for review and now an includes 'percentage' support:

## Match on list or single entry:
```
role = admin
preferredLang = en,es
```
will match context of `role=admin,preferredLang=en`

## Allow in 50% of the English and Spanish speakers
```
userId:percentage = 50
preferredLang = en,es
```

## Enable for 20% of the team admins
```
teamId:percentage = 20
role = admin
```

It doesn't address the case if a person is on multiple teams, but that specific of a context would require a custom strategy.

ivarconr:
Hi,

this is a clever way to use Unleash. We have discussed it internally and feel it is solving the same needs as strategy constraints. We have quite a few plans for strategy constraints, and how we can build other features on top of that concept. We are even discussing on Open-Sourcing part of that, to make it available for everyone. 

The matching strategy is a clever approach, but officially supporting this would also require us to take it to all the official SDKs. This would take effort and drives or focus in the wrong direction. 

For these reasons we have decided to not approve this PR. We are positive to approving #127, and you are free to open source the matching strategy, if you think this can be valuable to more.

checketts:
Thanks for the update. I definitely don't want this feature to distract from the current Unleash direction. I don't have a good feel for all the constraint capabilities, but with the roadmap I noted the idea of reusable/shared config between toggles and that definitely would be interesting to use.

I would love to have the constraint feature to be open sourced, just to ensure that open source Unleash is as flexible as it needs to be. 

There is one case that I would like to bring up to re-consider: 

Would there be interest in including something like this in the Java SDK, without registering it as strategy, but making it a base for other strategies?  

If that sounds interesting I would refactor the existing strategies to leverage this one, effectively sharing the implementation, and simplifying the codebase, but it would allow others users to extend it as well if they wish to create a fallback?

Issue #129, opened 2021-03-16
feat: Support warm-up state from "location"


ivarconr:
I some scenarios one could risk that the unleash-api is unavailable on startup (networks are unreliable after all). And relying on the built in backup file could not be an option in the world of immutable (docker) containers. 

It could be good to allow the user to specify the initial state to use (instead of the built-in backup-file location) where the user have curated the latest feature toggle configuration snapshot to use. This should be a pure dump from the `/api/client/features` endpoint. The dump could be produced at the CI server when the application is bundled or later through the infrastructure (e.g. config maps or similar). 

We would (probably?) not want to read the temporary backup if such "hot backup location" is defined. The location could either be a absolute file (in the docker container) or a URL (e.g. a static file server, S3 or simial).

checketts:
If a warm-up state exists, then it seems like the temporary backup would still be preferred (if it has content) since it would have been created after the warm-up state. Am I misunderstanding a part of that?

ivarconr:
I think it depends. Say the warm-up location is a file on S3 which is updated regularly (or as a result of the webhook), then it would probably be as fresh as it gets? 

But yes, I don't feel there is a clear answer, but given the temp backup file exists and the application has been running recently, it is probably the freshest state in most situations.

PR #131, opened 2021-03-16
Add a FeatureToggle bootstrapper.

chriswk:
Since our normal backup file location must be writable since we write back. 
This adds support for reading toggles from a place where Unleash does not have write access.
First implementation only adds support for accepting absolute paths to filesystem, but the interface should be extensible enough so we can add `s3` or `cloud-storage` support, though we probably want to keep that out of client-core, to avoid dragging in aws or google dependencies

fixes: #129


checketts:
It would be nice if the default absolute filesystem path could also support resources on the classpath. That way a resource that was compiled into a 'fat jar' would also be easily usable.

chriswk:
Good point, I'll add support for `classpath:` resources with this.

I do seem to remember that fat jar resources came with some extra configuration issues since the embedded file most likely is zipped inside a jar. I'll start by making sure we can find resources from classpath.

README.md
```diff
@@ -221,6 +221,14 @@ based on the values stored in `unleash-repo.json`. As a result of this, the seco
 * When `unleash-repo.json` does not exists
 * When the named feature toggle does not exist in `unleash-repo.json`
 
+## Bootstrapping
+Unleash supports bootstrapping from a local file where Unleash does not have write access.
```
chriswk, line 225:
```suggestion
Unleash supports bootstrapping from a local file without need for write access.
```

README.md
```diff
@@ -221,6 +221,14 @@ based on the values stored in `unleash-repo.json`. As a result of this, the seco
...
+Useful for applications deployed to ephemeral containers (docker) or more locked down file systems where Unleash's need to write back to the backup file is not desired.
+Configure using
+* `UNLEASH_BOOTSTRAP_FILE` - path to file in toggle json format as defined by the Unleash servers `/api/client/features` endpoint
+* `UNLEASH_BOOTSRAP_FILE_CHECKSUM`
```
chriswk, line 229:
```suggestion
* `UNLEASH_BOOTSTRAP_FILE_CHECKSUM`
```

ivarconr:
> we can add s3 or cloud-storage support, though we probably want to keep that out of client-core, to avoid dragging in aws or google dependencies

I was thinking it might be enough to support fetching a json file from a URL, maybe with some authentication as part of that. 

Maybe you should be allowed to just push in an input stream? Then you can fetch it however you like, and you can just pass on the input stream as part of the configuration? Or would it be better to then say it must be a json string?

chriswk:
I think you're right. We should probably simplify the BootstrapHandler to accept a string which it parses to a set of toggles. Then it would be up to the user to figure out how they want to provide the String. Seeing as how the current FileHandler only uses Java core APIs providing it could be a starter for how to write a ToggleBootstrapHandler. I'll have a look at changing the API a bit

src/main/java/no/finn/unleash/repository/FeatureToggleRepository.java
```diff
@@ -40,7 +39,14 @@ public FeatureToggleRepository(
         this.toggleFetcher = toggleFetcher;
         this.eventDispatcher = new EventDispatcher(unleashConfig);
 
-        toggleCollection = toggleBackupHandler.read();
+        this.toggleCollection = toggleBackupHandler.read();
```
ivarconr, line 42: So we are 100% sure toggleBackupHandler cannot return null?

src/main/java/no/finn/unleash/repository/FeatureToggleRepository.java
```diff
@@ -40,7 +39,14 @@ public FeatureToggleRepository(
...
+        if (toggleCollection != null && toggleCollection.getFeatures().isEmpty()) {
+            if (unleashConfig.getToggleBootstrapProvider() != null) {
+                toggleCollection =
+                        new ToggleBootstrapHandler(unleashConfig)
```
ivarconr, line 46: I feel it would be more readable if we already have constructed the `ToggleBootstrapHandler` even though it wont be needed if a `ToggleBootstrapProvider` is not provided.

ivarconr:
I really like the inversion of control where the user now can just specify a `string` producer. This means the user can read their warmup state from anywhere (url, redis, S3, memcached, etc) without injecting support for everything in to the SDK. 

Would be nice to see a Howto guide at some point where you curate a hot file in a S3 bucket and read it at startup.

chriswk:
I can look into adding a how-to for that

src/main/java/no/finn/unleash/repository/FeatureToggleRepository.java
chriswk: I was trying to avoid unnecessary initialization, but I'd be fine with just instantiating the BootstrapHandler and let it handle whether or not its provider is initialized. I'll change the code.

src/main/java/no/finn/unleash/repository/FeatureToggleRepository.java
chriswk: It can absolutely return null, I've annotated the backupHandler method to signify this. Just had quite a few tests fail because I removed the check on L43 as to whether or not it was null.

chriswk: Though that does seem to be due to the fact that we mock the entire class, Looking at the code for BackupHandler it seems to always return a ToggleCollection, either with feature toggles or with an empty feature toggle list.

PR #135, opened 2021-04-20
Add support for leveraging namePrefix in Java SDK

checketts:
Also warn if trying to check for toggles that don't match the prefix to avoid confusion for why it is always disabled.

### DIFF
--- a/src/main/java/no/finn/unleash/repository/ToggleBootstrapHandler.java
+++ b/src/main/java/no/finn/unleash/repository/ToggleBootstrapHandler.java
@@ -43,7 +43,7 @@ public class ToggleBootstrapHandler {
         return new ToggleCollection(Collections.emptyList());
     }
 
-    private static class ToggleBootstrapRead implements UnleashEvent {
+    public static class ToggleBootstrapRead implements UnleashEvent {
         private final ToggleCollection toggleCollection;
 
         private ToggleBootstrapRead(ToggleCollection toggleCollection) {
--- a/src/main/java/no/finn/unleash/util/UnleashURLs.java
+++ b/src/main/java/no/finn/unleash/util/UnleashURLs.java
@@ -1,7 +1,10 @@
 package no.finn.unleash.util;
 
+import no.finn.unleash.lang.Nullable;
+
 import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 
 public class UnleashURLs {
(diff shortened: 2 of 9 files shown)


src/test/java/no/finn/unleash/event/SubscriberTest.java
```diff
@@ -61,15 +65,14 @@ void subscriberAreNotified() {
...
-        assertThat(testSubscriber.events)
-                .hasSize(
-                        1 // Bootstrapping
-                                + 3 // feature evaluations
-                                + 2 // toggle fetches
-                                + 1 // unleash ready
-                                + 1 // client registration
-                                + 1 // client metrics
-                        );
+//        assertThat(testSubscriber.events).filteredOn(e -> e instanceof ToggleBootstrapHandler.ToggleBootstrapRead).hasSize(1);
```
checketts, line 68: I had to update this test to get it passing (it was failing in master already)

src/main/java/no/finn/unleash/repository/ToggleBootstrapHandler.java
```diff
@@ -43,7 +43,7 @@ public ToggleCollection read() {
         return new ToggleCollection(Collections.emptyList());
     }
 
-    private static class ToggleBootstrapRead implements UnleashEvent {
+    public static class ToggleBootstrapRead implements UnleashEvent {
```
checketts, line 46: I had to make this public so the event could be introspected in the test (I can't imagine we want the event to not be visible to subscribers.)

ivarconr:
Read quickly through, and the only concern i have is if the implementation will allow multiple query fields at the same time (e.g. project and namsPrefix)?

checketts:
This implementation does not support that. I was just reading through the node code and realized that was supported, so I'll update this PR to support that as well.

checketts:
I've updated the code to support both project and namePrefix now.

checketts:
I just added the changelog entry. Sorry I had forgotten that before!

src/main/java/no/finn/unleash/util/UnleashURLs.java
```diff
@@ -25,22 +28,35 @@ public URL getFetchTogglesURL() {
...
+        appendParam(suffix, "namePrefix", namePrefix);
+
+        try {
+            return URI.create(fetchTogglesURL + suffix.toString()).normalize().toURL();
+        } catch (IllegalArgumentException|MalformedURLException e) {
+            throw new IllegalArgumentException("fetchTogglesURL [" + fetchTogglesURL + suffix + "] was not URL friendly.", e);
+        }
+    }
+
+    private void appendParam(StringBuilder suffix, String key, @Nullable String value) {
```
ivarconr, line 51: @chriswk will this also be usable for adding array of tags later on?

ivarconr (approved):
Thanks for yet another great contribution!

LGTM!

Issue #138, opened 2021-06-07
Move package to io.getunleash


chriswk:
We should move the library away from the no.finn umbrella now that we've acquired the io.getunleash groupId.

checketts:
This kind of a change is very disruptive for users. For the most used classes (and interfaces), I recommend duplicating the class, and having the old class sub-class the new class. The allows a transition.

Example:

`no.finn.unleash.Unleash extends io.getunlesh.unleash.Unleash`

PR #139, opened 2021-06-07
Feat/custom stickiness

chriswk:


chriswk:
Hmm, just realised I did the refactoring of package names here as well. I think if you look at the diff for ed15b43 that should be the custom stickiness changes.

src/main/java/io/getunleash/variant/VariantUtil.java
```diff
@@ -61,6 +62,17 @@ private static String getIdentifier(UnleashContext context) {
                                                 .orElse(Double.toString(Math.random()))));
```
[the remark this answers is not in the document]
ivarconr: ooh, nice catch! no reason not to use SecureRandom

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -1,27 +1,27 @@
-package no.finn.unleash;
+package io.getunleash;
```
checketts, line 1: Don't you want to leave the old package in a deprecated state for a few releases? At lease for the classes that are most likely to be used, like `DefaulUnleash`?

chriswk: The plan is to make a major release with the package moved, but good point we should probably have the old package in place with a deprecation warning and just a redirect for the typical initialization stories. I'll update the PR.

chriswk: Since we're relocating the maven package coords as well, I think it's reasonable to expect our users to do a query-replace for their package imports. I'll remove the package move from this PR and do that in separate PR before we release 5.0 of the client. The stickiness does not require a new major version.

checketts: The main problem comes from intermediate usages. For example if my company creates an internal helper library called helper:v1 that uses Unleash:v3 then it locks the app into not being able to upgrade their Unleash version to the next version until the helper library is upgraded.

Another example is with the recently merged in Chaos Monkey for Spring Boot support for Unleash.

checketts: Once work around is the users could put both artifacts on the classpath and instantiate them twice. It would be good to avoid making users do that though

chriswk: Good point on internal helper libraries, we'll have a discussion for a proper package migration story then. 

Personally I would be fine with not being able to upgrade the backing library for a helper library until the helper library has upgraded, but I can see how that would impede progress/process as well. And I do agree with trying to avoid having our users double instantiate the UnleashClient.

I've removed the package migration away from this issue, so it's moot for this PR. But open for ideas over on #138

src/main/java/no/finn/unleash/variant/VariantUtil.java
```diff
@@ -61,6 +62,17 @@ private static String getIdentifier(UnleashContext context) {
                                                 .orElse(Double.toString(Math.random()))));
     }
 
+    private static String randomString() {
+        int randSeed = new Random().nextInt(100000);
```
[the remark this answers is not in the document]
chriswk: @muse-dev ignore

PR #145, opened 2021-11-16
task: Add OkHttpToggleFetcher implementation

chriswk:
This PR adds a new ToggleFetcher using Square's OkHttp library to perform the HTTP calls, rather than using java.net's internal HttpClient.

Since OkHttp handles caching and redirect logic for use, it reduces logic we have to worry about, at the cost of adding a dependency that does bring with it some additional transitive dependencies.

This PR is not feature complete, because we will have to make sure the OkHttp client can also be convinced to use the proxy logic that has been written for the current java.net implementation. Copying all mocks and tests for the current HttpToggleFetcher and switching to using the OkHttpToggleFetcher did reveal a couple of differences that will need to be verified against our server.

One is that OkHttp apparently suffixes ETags with --gzip if it receives a gzip response; we'll have to make sure that our Unleash server still sees this as the same ETag, and if not, configure OkHttp to add the --gzip suffix.

The other is that apparently HttpUrl's toString() methods suffix toplevel URLs with `/` whereas the old code took the response header raw (without a trailing `/`). Semantically it's the same, and the redirect request does actually happen, so I'm pretty sure it confirms that we still do the same logic for a 308.

Created as Draft for now, open for discussion


pom.xml
```diff
@@ -56,9 +57,13 @@
         <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
-            <version>2.8.8</version>
+            <version>2.8.9</version>
+        </dependency>
+        <dependency>
+          <groupId>com.squareup.okhttp3</groupId>
+          <artifactId>okhttp</artifactId>
+          <version>${version.okhttp}</version>
```
checketts, line 65: Make the dependency `<optional>true</optional>` to allow it to be included and used if the user also include the OKHttp library.

(I'm interested in creating one with Spring's WebClient as well, since I'll get success/failure and timing metrics)

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -50,7 +51,9 @@
     private static FeatureToggleRepository defaultToggleRepository(UnleashConfig unleashConfig) {
         return new FeatureToggleRepository(
                 unleashConfig,
-                new HttpToggleFetcher(unleashConfig),
+                unleashConfig.isUseOkHttpClient()
```
checketts, line 54: Instead of a boolean, it would be nice of unleashConfig to allow providing a fetcher implementation:

```
unleashConfig.toggleFetcher(new OkHttpToggleFetcher(unleashConfig))
```

Though that does mean the the unleash config being passed in is being created when the fetcher is being constructed.

checketts: To work around the config being created when the instantiation happens, we should make it a factory:

```java
unleashConfig.toggleFetcherFactory(config -> new OkHttpToggleFetcher(config))
```

pom.xml
chriswk: Thanks, good point :)

src/main/java/io/getunleash/DefaultUnleash.java
chriswk: Thanks Clint, another good point :+1: I'll update the PR.

pom.xml
chriswk: done, I'm considering allowing you to configure the OkHttpClient directly, so you can setup an interceptor for prometheus metrics for the client

andreas-unleash (approved):

chriswk:
@checketts - Good question. Yes, you can write your own fetcher and supply that to the config builder.

src/main/java/io/getunleash/util/UnleashConfig.java
```diff
@@ -309,6 +327,7 @@ public static class Builder {
         private long fetchTogglesInterval = 10;
         private long sendMetricsInterval = 60;
         private boolean disableMetrics = false;
+        private UnleashFeatureFetcherFactory unleashFeatureFetcherFactory = HttpFeatureFetcher::new;
```
chriswk, line 330: @checketts Implement UnleashFeatureFetcherFactory and send it to the builder to use your custom Fetcher - `unleashFeatureFetcherFactory(..)` :)

ivarconr (approved):

PR #148, opened 2021-12-15
Upgrades log4j2 to 2.16.0

chriswk:
- Though we don't distribute log4j on runtime/compile classpath, since
  we have it as a test dependency, it's good to clear up confusion and
  make sure to use a presumably bugfree version for our test code as
  well


sighphyre (approved):
Yay! Nice to be ahead of this before someone comes asking

PR #154, opened 2022-04-11
docs: document all configuration options

### LINKED ISSUE #153: Add documentation for additional configuration parameters
I'd like to request more documentation on what customizable methods are available when configuring Unleash. For example, I didn't realize we had an option to change how often we are polling Unleash for updates until @ivarconr pointed out the `fetchTogglesInterval` configuration to me in https://github.com/Unleash/unleash-client-java/blob/main/src/main/java/io/getunleash/util/UnleashConfig.java#L376

thomasheartman:
This PR adds a table listing all `UnleashConfig` builder methods. The table also adds descriptions, default values, and indicates whether the options must be configured or not.

Addresses and closes #153.

## Todo: verify this

Before this can go in, someone with more knowledge of the client should look through the descriptions and make sure they're correct. There's also a number of options that I don't know what to write about, so I'll need some input there. Particular configuration options will be called out inline.


README.md
```diff
@@ -343,5 +352,33 @@ The generated report will be available at ```target/site/jacoco/index.html```
...
+
+| Method name                                | Description                                                                                                                                                         | Required | Default value                        |
+|--------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------|--------------------------------------|
+| `appName`                                  | The name of the application as shown in the Unleash UI. Registered applications are listed on the Applications page.                                                | Yes      | `null`                               |
+| `backupFile`                               | The path to the file where [local backups](#local-backup) get stored.                                                                                               | No       | `"java.io.tmpdir/unleash-repo.json"` |
```
thomasheartman, line 367: Is this the correct default backup file location?

README.md
```diff
@@ -343,5 +352,33 @@ The generated report will be available at ```target/site/jacoco/index.html```
...
+| `customHttpHeader`                         | Add a [custom HTTP header](#custom-http-headers) to the list of HTTP headers that will the client sends to the Unleash API. Each method call will add a new header. | No       | N/A                                  |
```
thomasheartman, line 368: Presumably, this isn't a required setting _per se_. However, this seems to be the only way to set the authentication token, and the client won't work without that, right?

README.md
```diff
@@ -343,5 +352,33 @@ The generated report will be available at ```target/site/jacoco/index.html```
...
+| `disableMetrics`                           | A boolean indicating whether the client should disable sending usage metrics to the Unleash server.                                                                 | No       | `false`                              |
+| `enableProxyAuthenticationByJvmProperties` | Enable support for [using JVM properties for HTTP proxy authentication](#http-proxy-with-authentication).                                                           | No       | `false`                              |
+| `environment`                              | The name of the current environment.                                                                                                                                | No       | `null`                               |
+| `fallbackStrategy`                         | ???                                                                                                                                                                 | No       | `null`                               |
```
thomasheartman, line 373: What is this? Fallback strategy for what?

README.md
```diff
@@ -343,5 +352,33 @@ The generated report will be available at ```target/site/jacoco/index.html```
...
+| `fetchTogglesInterval`                     | How often (in seconds) the client should check for toggle updates.                                                                                                  | No       | `10`                                 |
+| `instanceId`                               | An identifier for your instance. Typically a hostname, pod id or something similar. Unleash uses this internally for ...something                                   | Yes      | `null`                               |
```
thomasheartman, line 375: What _does_ Unleash do with the instance ID? Is it surfaced somewhere?

README.md
```diff
@@ -343,5 +352,33 @@ The generated report will be available at ```target/site/jacoco/index.html```
...
+| `namePrefix`                               | ???                                                                                                                                                                 | No       |                                      |
```
thomasheartman, line 376: What is a name prefix? Is it a prefix applied to all toggles?

README.md
```diff
@@ -343,5 +352,33 @@ The generated report will be available at ```target/site/jacoco/index.html```
...
+| `projectName`                              | The name of the Unleash project?                                                                                                                                    | No       |                                      |
```
thomasheartman, line 377: This probably isn't the Unleash project name, but instead ... something Java-related?

README.md
```diff
@@ -343,5 +352,33 @@ The generated report will be available at ```target/site/jacoco/index.html```
...
+| `proxy`                                    | ???                                                                                                                                                                 | No       |                                      |
```
thomasheartman, line 378: Is this the HTTP proxy that's referenced other places in the readme? What is that?

README.md
```diff
@@ -343,5 +352,33 @@ The generated report will be available at ```target/site/jacoco/index.html```
...
+| `scheduledExecutor`                        | ???                                                                                                                                                                 | No       |                                      |
```
thomasheartman, line 379: What does this do?

README.md
sighphyre: I don't think so, the backup file name contains the appName as well, from the source code (tweaked for clarity): 
```
System.getProperty("java.io.tmpdir") + File.separatorChar + "unleash-" + appName + "-repo.json";
```

README.md
sighphyre: I can see a use case here where someone turns off Auth on the Unleash server side. A few of the SDKs have this as an optional property. Might be worth a warning that for the overwhelming majority of cases this is required to make the system work though!

README.md
sighphyre: Looks like this is used when checking the strategies for a toggle. When the SDK tries to lookup the strategy for a given toggle it uses this as a fallback if the strategy can't be mapped to one of the SDK strategy types (guessing this is for a usecase where you may be running an older SDK with a newer Unleash with new strategies)

README.md
sighphyre: This nudges the SDK to only return toggles that start with a given prefix so if you set this to:

'coolFeature'

And your toggles server side are called:

'notAcoolFeature'
'coolFeature42'

You'll only get the second one back in the SDK call. This supported on the Unleash server but only the Node and Java SDKs make use of it (insofar as I'm aware and confirmed by our compat table)

README.md
sighphyre: It is the Unleash project name! Related to the one above, this nudges Unleash to only query toggles in the given project. You should be able to do the same thing with an appropriate token constrained to only that project (this might be if you want to turn off Auth?)

README.md
sighphyre: This is an object containing proxy details in case there's a third party proxy between your SDK and your Unleash server

README.md
sighphyre: This beastie controls the timing and running of tasks (mostly for fetching Toggles), there's a default threaded implementation called `UnleashScheduledExecutorImpl`, people who know this code better than me will have to clue me in on when/why you'd want your own custom implementation here though

sighphyre (approved):
LGTM

PR #157, opened 2022-05-18
feat: add support for global segments

andreas-unleash:
Global segment implementation
Segments allow for reusable constraint chains.

## About the changes

Closes # .

### Important files

To accommodate the changes and receive the segments, the following files are Deprecated
HttpToggleFetcher -> Replaced with HttpFeatureFetcher
FeatureToggleRepository -> Replaced with FeatureRepository
JsonToggleCollectionDeserializer -> Replaced with JsonFeaturesDeserializer
ToggleBackupHandlerFile -> Replaced with FeatureBackupHandlerFile
ToggleBootstrapHandler -> Replaced with FeatureBootstrapHandler


src/main/java/io/getunleash/Segment.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+
+public class Segment {
+    private int id;
+    private String name;
+    @Nullable
+    private String description;
+    private List<Constraint> constraints;
+    private String createdBy;
+    private String createdAt;
```
chriswk, line 14: I'm not sure we need description, createdBy and createdAt, they may be part of the body, but it's not really needed for performing the constraint matching, maybe drop down to just `id, name, constraints`?

I think we've already configured Jackson to parse without failing on unknown fields, so it should be legal

src/main/java/io/getunleash/event/UnleashSubscriber.java
```diff
@@ -31,4 +32,10 @@ default void togglesBackedUp(ToggleCollection toggleCollection) {}
     default void toggleBackupRestored(ToggleCollection toggleCollection) {}
 
     default void togglesBootstrapped(ToggleCollection toggleCollection) {}
+
+    default void featuresBootstrapped(FeatureCollection featureCollection) {}
+
+    default void featuresBackedUp(FeatureCollection featureCollection) {}
+
+    default void featuresBackupRestored(FeatureCollection featureCollection) {}
 }
```
chriswk, line 41: Cool to have these :)

src/main/java/io/getunleash/util/UnleashURLs.java
```diff
@@ -10,12 +10,15 @@ public class UnleashURLs {
...
 
+    private final URL featuresURL;
+
     public UnleashURLs(URI unleashAPI) {
         try {
             String unleashAPIstr = unleashAPI.toString();
             fetchTogglesURL = URI.create(unleashAPIstr + "/client/features").normalize().toURL();
             clientMetricsURL = URI.create(unleashAPIstr + "/client/metrics").normalize().toURL();
             clientRegisterURL = URI.create(unleashAPIstr + "/client/register").normalize().toURL();
+            featuresURL = URI.create(unleashAPIstr + "/client/register").normalize().toURL();
```
chriswk, line 21: Do we need another URL here? Isn't the new data already coming from the fetchTogglesURL?

Small nit, but an URL named featuresURL that's pointing to a register endpoint seems like a smell.

chriswk (commented):
Some comments, and as mentioned in our VC, I'm missing some tests, but you've already confirmed you're going to add those, so am looking forward to seeing this get merged to main :)

src/main/java/io/getunleash/repository/HttpFeatureFetcher.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    private ClientFeaturesResponse followRedirect(HttpURLConnection request) throws IOException {
+        String newUrl = getLocationHeader(request);
+
+        request = openConnection(new URL(newUrl));
+        request.connect();
+        LOG.info(
+                "Redirecting from {} to {}. Please consider to update your config.",
```
sighphyre, line 111:
```suggestion
                "Redirecting from {} to {}. Please consider updating your config.",
```

examples/cli-example/src/main/java/io/getunleash/example/AdvancedConstraints.java
```diff
@@ -8,8 +8,8 @@
 public class AdvancedConstraints {
     public static void main(String[] args) throws InterruptedException {
         UnleashConfig config = UnleashConfig.builder().appName("client-example.advanced")
-            .customHttpHeader("Authorization", "default:development.5ad8536b3f4f527355a92a628b4519751804d499ac7155794eb039bf")
-            .unleashAPI("https://app.unleash-hosted.com/hosted/api").instanceId("example")
+            .customHttpHeader("Authorization", "*:development.5f4b073484dd6de99fc2c8a9667b457e6cdb850224ffd71c32996aea")
```
sighphyre, line 11: Is this an intentional change? I ask because every time I work on an SDK I change this equivalent code for local dev purposes and often forget to roll it back.

src/main/java/io/getunleash/repository/JsonFeatureParser.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import java.io.Reader;
+
+final class JsonFeatureParser {
+
+    private JsonFeatureParser() {}
+
+    public static String toJsonString(FeatureCollection featureCollection) {
+        Gson gson = new GsonBuilder().create();
```
sighphyre, line 12: It's been a long time since I've worked with GSON but if memory serves, the GSON objects are both thread safe and expensive to instantiate. Might be worthwhile to make these once and hold them if possible

sighphyre (commented):
Overall I think this looks good. I've left some small comments lying around. My big question is: this looks like a large change that will affect the way the SDK is used, so I'm guessing this means a major release? If that's the case we probably need to also look at the docs for this but we probably need to handle that in a separate PR. I'd like @chriswk to take one last look if possible since my Java is very rusty

andreas-unleash:
I am not sure about the major release.  I think the way the sdk is interacted with (DefaultUnleash) - has not changed. Only the way the internals are managed has

src/main/java/io/getunleash/Segment.java
andreas-unleash: Removed

chriswk (approved):
I think we need to be very careful about throwing exceptions, but the rest of work looks really solid, so I'm :+1:.

chriswk:
I'm sorry for pinging you @checketts , but your reviews have been very valuable earlier, so if you have the time I'd greatly appreciate to have your opinion as well

checketts:
Glad to take a look. Is there a summary of the motivation or feature driving this change?

src/main/java/io/getunleash/repository/FeatureRepository.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        unleashConfig
+                .getScheduledExecutor()
+                .setInterval(updateFeatures(), 0, unleashConfig.getFetchTogglesInterval());
+    }
+
+    public static synchronized FeatureRepository getInstance() {
+        if (instance == null) {
+            throw new AssertionError("FeatureRepository:: You have to call init first");
```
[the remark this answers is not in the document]
checketts: Since the API already has a good fallback (evaluating features as false) when the Unleash API is not reachable, then I can see avoiding a 'fast fail' could be acceptable. 

I like fast failure so I can fix it quicker though. And not be scratching my head wondering why the toggles aren't working (and I've overlooked a single log message)

src/main/java/io/getunleash/repository/HttpToggleFetcher.java
```diff
@@ -13,6 +13,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+@Deprecated
```
checketts, line 16: Including a comment explaining what replaces the deprecated API is very helpful. That way the user has a hint of how to migrate.

src/main/java/io/getunleash/repository/SegmentCollection.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public final class SegmentCollection implements Serializable {
```
checketts, line 14: Why is this serializable?

src/main/java/io/getunleash/util/ArrayUtils.java
```diff
@@ -0,0 +1,11 @@
+package io.getunleash.util;
+
+import io.getunleash.lang.Nullable;
+import java.util.Collection;
+
+public class ArrayUtils {
```
checketts, line 6: Arrays aren't Collections. Should this be called `CollectionUtils`?

src/test/java/io/getunleash/UnleashTest.java
```diff
@@ -33,12 +37,23 @@ public void setup() {
...
                         .unleashAPI("http://localhost:4242/api/")
                         .environment("test")
+                        .scheduledExecutor(mock(UnleashScheduledExecutor.class))
                         .unleashContextProvider(contextProvider)
                         .build();
 
         unleash = new DefaultUnleash(config, toggleRepository, new UserWithIdStrategy());
     }
 
+    private void setMock(FeatureRepository mock) {
```
checketts, line 47: I'm not a huge fan of this black magic. I'm guessing it is needed for setting a static field? Perhaps a package private setter with a `@VisibleForTesting` annotation would be a little clearer.

src/test/java/io/getunleash/repository/HttpFeatureFetcherTest.java
```diff
@@ -0,0 +1,278 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+@ExtendWith(WireMockExtension.class)
+@WireMockSettings(failOnUnmatchedRequests = false)
+public class HttpFeatureFetcherTest {
+
+    @ConfigureWireMock Options options = wireMockConfig().dynamicPort();
+    @InjectServer WireMockServer serverMock;
```
checketts, line 30: Nice addition of wiremock. Been meaning to do this very thing for a while. Thanks!

src/main/java/io/getunleash/repository/FeatureRepository.java
```diff
@@ -0,0 +1,100 @@
...
+    private final FeatureBootstrapHandler featureBootstrapHandler;
+    private final HttpFeatureFetcher featureFetcher;
+    private final EventDispatcher eventDispatcher;
+
+    private FeatureCollection featureCollection;
+    private boolean ready;
+
+    private static FeatureRepository instance = null;
+
+    private FeatureRepository(UnleashConfig unleashConfig) {
```
checketts, line 25: Why the static `FeatureRepository` approach with private constructor? It forbids an app from pointing to 2 Unleash servers (admittedly an unusual use case, but not unreasonable)

The static instances being set below could just as easily be constructor arguments, which simplify the unit tests and just keep things from being fragile.

checketts (commented):
I had some questions. I'm still not 100% clear on what the segments feature is. I don't like the static initializers/instance field unless there is some specific usecase we are trying to support.

Very glad to see wiremock in the tests!

checketts:
Nice cleanup of the singletons and adding the deprecation comments. Thanks for doing that.

src/main/java/io/getunleash/repository/FeatureRepository.java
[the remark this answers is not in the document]
sighphyre: Mixed feelings on this one but I think this is okay. It's pretty easy to setup some of the SDKs wrong and then never get toggles resolving correctly, it's super frustrating when that happens and small comfort when I realise I'm being dumb. I think so long as the error can only happen during the setup code it's okay, it'd be a huge worry for me if the SDK could crash its host process later. I guess this is the choice between fail-fast or never-work-at-all

src/main/java/io/getunleash/metric/UnleashMetricsSender.java
```diff
@@ -46,7 +46,8 @@ public JsonElement serialize(
                 Type type,
                 JsonSerializationContext jsonSerializationContext) {
             return new JsonPrimitive(ISO_INSTANT.format(localDateTime.toInstant(ZoneOffset.UTC)));
-        };
+        }
```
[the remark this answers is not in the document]
sighphyre: Definitely doesn't need to be part of this PR! Just looked funny to my eyes and this was a friendly reminder to double check local tooling

src/main/java/io/getunleash/repository/SegmentCollection.java
ivarconr: /cc @andreas-unleash

src/main/java/io/getunleash/strategy/segments/Lazy.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public Lazy<Optional<T>> filter(Predicate<T> predicate) {
+        return new Lazy<>(() -> Optional.of(get()).filter(predicate));
+    }
+
+    public static <T> Lazy<T> of(Supplier<T> supplier) {
+        return new Lazy<>(supplier);
+    }
+}
```
ivarconr, line 46: I cannot immediately see where this class is used @andreas-unleash ?

src/main/java/io/getunleash/util/CollectionUtils.java
```diff
@@ -0,0 +1,11 @@
...
+
+import io.getunleash.lang.Nullable;
+import java.util.Collection;
+
+public class CollectionUtils {
+
+    public static boolean isNullOrEmpty(@Nullable Collection<?> collection) {
+        return collection == null || collection.isEmpty();
+    }
+}
```
ivarconr, line 11: Where is this used?

src/main/java/io/getunleash/repository/ClientFeaturesResponse.java
```diff
@@ -0,0 +1,65 @@
+package io.getunleash.repository;
+
+import io.getunleash.lang.Nullable;
+import java.io.Serializable;
```
ivarconr, line 4: unused import.

src/main/java/io/getunleash/repository/ClientFeaturesResponse.java
```diff
@@ -0,0 +1,65 @@
+package io.getunleash.repository;
+
+import io.getunleash.lang.Nullable;
+
+import java.util.Collections;
+
+public final class ClientFeaturesResponse extends FeatureToggleResponse {
```
ivarconr, line 7: I struggle a bit to understand why we would want this wrapper around "FeatureToggleResponse". Why can't we simply move the segment collection in to the FeatureToggleResponse implementation and avoid this extra layer of inheritance?

andreas-unleash: I followed the same practice (and class name) as in the node sdk.  Nothing prevents us from using a modified version of the FeatureToggleResponse. imo inheritance is better in this situation, but happy to change if you disagree

ivarconr: > imo inheritance is better in this situation

why do you prefer inheritance in this situation?

src/main/java/io/getunleash/repository/SegmentCollection.java
ivarconr: I assume it needs to be in order to be able to write it to the file backup. (or gson supports anything..)

src/main/java/io/getunleash/repository/SegmentCollection.java
```diff
@@ -0,0 +1,44 @@
...
+
+    static final long serialVersionUID = 1214L;
+    private final Collection<Segment> segments;
+    private final transient Map<Integer, Segment> cache;
```
ivarconr, line 18: nice. Are we sure this is correctly re-populated when segments are read from the backup file?

src/main/java/io/getunleash/repository/BackupHandler.java
```diff
@@ -0,0 +1,7 @@
+package io.getunleash.repository;
```
ivarconr, line 1: I like this abstraction, but I would prefer general cleanup to happen in separate PRs as the number of changes in this PR gets a bit on the higher end in order to provide accurate feedback.

src/main/java/io/getunleash/repository/JsonFeaturesDeserializer.java
```diff
@@ -0,0 +1,45 @@
...
+        JsonObject root = rootElement.getAsJsonObject();
+        JsonArray segmentArray = root.getAsJsonArray("segments");
+        JsonArray togglesArray = root.getAsJsonArray("features");
+
+        Collection<FeatureToggle> toggles =
+                context.deserialize(togglesArray, TOGGLE_COLLECTION_TYPE);
+        Collection<Segment> segments = context.deserialize(segmentArray, SEGMENT_COLLECTION_TYPE);
+        return new FeatureCollection(
+                new ToggleCollection(toggles),
+                new SegmentCollection(segments != null ? segments : Collections.emptyList()));
```
ivarconr, line 43: would it make sense to move the null handling in to the SegmentCollection?

ivarconr (commented):
I would love to see a couple of tests related to the new segmentation capability. E.g. inside "DefaultUnleash" I would assume to see a few test for a toggle with a somewhat complex segment that will evaluate to true or false depending on the context fields provided.

src/main/java/io/getunleash/repository/SegmentCollection.java
andreas-unleash: I will write a test for it

src/main/java/io/getunleash/repository/JsonFeaturesDeserializer.java
andreas-unleash: yes it would, thanks for the heads up

src/main/java/io/getunleash/util/UnleashURLs.java
ivarconr: please remove this URL before merging @andreas-unleash

ivarconr: my bad, it was removed 👍🏼

src/main/java/io/getunleash/repository/ClientFeaturesResponse.java
andreas-unleash: It allows me to make use of the previous functionality, taking away the risk of breaking something existing, 
and the response class matches that of the node sdk.  
Nothing. stopping us from only modifying existing classes to achieve the same results, but I chose to inherit

ivarconr (commented):
It would be good if you can update client-specification to v4.2.0 ([here](https://github.com/Unleash/unleash-client-java/blob/28eae972a742dbf95025404cd11a368f21653be7/pom.xml#L13)) and run the integration tests. It seems to be one test failing.

andreas-unleash:
@ivarconr updated and fixed

ivarconr (approved):

chriswk:
[Java Segmentation SDK PoC](https://trello.com/c/6jgEzGiO/1209-java-segmentation-sdk-poc)

PR #160, opened 2022-06-28
Add getFeatureToggleDefinition to interface

Chaosmi:
## About the changes
Adds `getFeatureToggleDefinition` to the interface so we don't need to refer to the concrete `DefaultUnleash` when we want to use that method.

The method was initially added in https://github.com/Unleash/unleash-client-java/pull/36

## Discussion points
Please let me know if there are any outstanding reasons this method should not be on the interface or if we'd like to expose the functionality in a different way. Thanks

### DIFF
--- a/src/main/java/io/getunleash/DefaultUnleash.java
+++ b/src/main/java/io/getunleash/DefaultUnleash.java
@@ -190,6 +190,12 @@ public class DefaultUnleash implements Unleash {
         return getVariant(toggleName, contextProvider.getContext(), defaultValue);
     }
 
+    /**
+     * Use more().getFeatureToggleDefinition() instead
+     *
+     * @return the feature toggle
+     */
+    @Deprecated
     public Optional<FeatureToggle> getFeatureToggleDefinition(String toggleName) {
         return ofNullable(featureRepository.getToggle(toggleName));
     }
@@ -243,6 +249,11 @@ public class DefaultUnleash implements Unleash {
         @Override
         public List<String> getFeatureToggleNames() {
             return featureRepository.getFeatureNames();
+        }
+
+        @Override
+        public Optional<FeatureToggle> getFeatureToggleDefinition(String toggleName) {
+            return ofNullable(featureRepository.getToggle(toggleName));
         }
 
         @Override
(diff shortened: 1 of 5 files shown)


src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -190,6 +190,7 @@ public Variant getVariant(String toggleName, Variant defaultValue) {
         return getVariant(toggleName, contextProvider.getContext(), defaultValue);
     }
 
+    @Override
     public Optional<FeatureToggle> getFeatureToggleDefinition(String toggleName) {
         return ofNullable(featureRepository.getToggle(toggleName));
     }
```
andreas-unleash, line 196: This seems to me better placed in the MoreOperations interface (like `getFeatureToggleNames` - notice how the one in Unleash interface was deprecated) - You could then access it with `more().getFeatureToggleDefinition('some-feature')`

src/test/java/io/getunleash/UnleashTest.java
```diff
@@ -225,8 +234,7 @@ public void should_return_empty_for_unknown_feature_toggle_definition() {
         FeatureToggle featureToggle = new FeatureToggle("test", false, asList(strategy1));
```
andreas-unleash, line 234: Nice

andreas-unleash (changes requested):

src/main/java/io/getunleash/DefaultUnleash.java
Chaosmi: Hey, thanks. that sounds cool. 

What implementation would you expect in `FakeUnleash.FakeMore`?
should we just throw a UnsupportedOperation exception for now or do we need to change this to return a dummy feature toggle?

Chaosmi: Thanks @andreas-unleash. I have now moved the method to the `MoreOperations` interface and changed `features` to be a map of `FeatureToggle` rather than a map of booleans.

andreas-unleash: Thanks @Chaosmi ! Approved

Issue #163, opened 2022-07-17
Make client timeouts for metrics sending configurable


chriswk:
### Describe the feature request

I wish for socket and read timeouts to be configurable for the http client I'm using.

### Background

We've just had a customer wondering about the clientTimeout for our metrics posting, and I've found that it's hard-coded to 1000 milliseconds. For customers with heavy load on their database posting metrics might take longer than this, so we should make it possible to configure the timeouts for our http client usage.

telekom-digioss:
Would it be possible to extend the default value for request timeout from 1000 to 5000 ms also ?
... all the other hardcoded timeouts have currently already 10000 ms. Only the metrics post has this low timeout where we struggle with

chriswk:
Hi Robert. I'll put the default to 10 seconds. PR incoming.

PR #164, opened 2022-07-18
task: Make http agent timeouts configurable

chriswk:
This PR makes connect and read timeouts for our http requests configurable, previously they've been hardcoded.
fixes: #163


sighphyre (approved):
LGTM

PR #165, opened 2022-07-22
task: Add an OkHttpMetricsSender

chriswk:
So, having added an OkHttpFeatureFetcher in a previous PR, this PR adds a metricssender which uses OkHttp.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ If you want to use OkHttp instead of HttpURLConnection you'll need a dependency 
 <dependency>
     <groupId>com.squareup.okhttp3</groupId>
     <artifactId>okhttp</artifactId>
-    <version>4.9+</version>
+    <version>4.10+</version>
 </dependency>
 ```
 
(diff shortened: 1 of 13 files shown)


thomasheartman:
@chriswk This PR has been marked as stale: do we still want to get it in or should we close it?

src/main/java/io/getunleash/metric/JavaHttpMetricsSender.java
```diff
@@ -25,7 +23,7 @@ public class UnleashMetricsSender {
     private final URL clientRegistrationURL;
     private final URL clientMetricsURL;
 
-    public UnleashMetricsSender(UnleashConfig unleashConfig) {
+    public JavaHttpMetricsSender(UnleashConfig unleashConfig) {
```
sighphyre, line 26: Very optional nit: Happy with the name here? The Java feels a bit redundant in a Java program. DefaultMetricsSender maybe?

sighphyre (approved):
Looks good, I've put down an optional thing around naming but I'm not going to block this if you're happy with the name. Let's get this beast in!

src/main/java/io/getunleash/metric/JavaHttpMetricsSender.java
chriswk: Well, it is using java.net.http :P But agreed that it feels redundant. Default is fair. I'll do the rename

PR #166, opened 2022-07-29
feat: Add possibility to disable polling

chriswk:
So, there are situations where you don't necessarily want to poll, so taking inspiration from our already existing `disableMetrics()` flag on UnleashConfig, this PR adds a `disablePolling()` flag which stops the repository from adding a task to our scheduled executor. 

So if you don't use `synchronousFetchOnInitialization()` and have no bootstrap, it is possible to have a situation where you don't have any features.

If you want to fetch once and then stop, set `fetchTogglesInterval` to 0, that will cause the executor to be set to `runOnce` instead of at a `fixedRate`


README.md
```diff
@@ -392,11 +393,12 @@ The `UnleashConfigBuilder` class (created via `UnleashConfig.builder()`) exposes
...
 | `customHttpHeader`                         | Add a [custom HTTP header](#custom-http-headers) to the list of HTTP headers that will the client sends to the Unleash API. Each method call will add a new header. Note: in most cases, you'll need to use this method to provide an API token. | No       | N/A                                                                                                                  |
 | `customHttpHeadersProvider`                | Add a custom HTTP header provider. Useful for [dynamic custom HTTP headers](#dynamic-custom-http-headers).                                                                                                                                       | No       | `null`                                                                                                               |
+| `disablePolling`                           | A boolean indicating whether the client should poll the unleash api for updates to toggles                                                                                                                                                       |
```
sighphyre, line 396: Absolute nit but everything else in this table seems to terminate in a period

```suggestion
| `disablePolling`                           | A boolean indicating whether the client should poll the unleash api for updates to toggles                                                                                                                                                       |
```
```suggestion
| `disablePolling`                           | A boolean indicating whether the client should poll the unleash api for updates to toggles.                                                                                                                                                       |
```

README.md
```diff
@@ -392,11 +393,12 @@ The `UnleashConfigBuilder` class (created via `UnleashConfig.builder()`) exposes
...
-| `fetchTogglesInterval`                     | How often (in seconds) the client should check for toggle updates.                                                                                                                                                                               | No       | `10`                                                                                                                 |
+| `fetchTogglesInterval`                     | How often (in seconds) the client should check for toggle updates. Set to `0` if you want to only check once                                                                                                                                     | No       | `10`                                                                                                                 |
```
sighphyre, line 401:
```suggestion
| `fetchTogglesInterval`                     | How often (in seconds) the client should check for toggle updates. Set to `0` if you want to only check once.                                                                                                                                     | No       | `10`                                                                                                                 |
```

src/main/java/io/getunleash/util/UnleashConfig.java
```diff
@@ -567,6 +589,16 @@ private String getBackupFile() {
             }
         }
 
+        /**
+         * Adds a custom http header for authorizing the client
+         * @param apiKey the client key to use to connect to the Unleash Server
+         * @return
+         */
+        public Builder apiKey(String apiKey) {
+            this.customHttpHeaders.put("Authorization", apiKey);
```
sighphyre, line 598: Nice, this is so much cleaner! Looks like this should also work on top of custom header providers without problems. 

Right now the examples on the README around headers still show the old usage where you'd need to inject the auth header manually, should we also update those to reflect this?

sighphyre (approved):
LGTM, few small suggestions but nothing blocking

thomasheartman:
@chriswk @sighphyre What's the state here? Doesn't seem like anything's happened for a while. Is this still being worked on?

src/main/java/io/getunleash/util/UnleashConfig.java
chriswk: Feels like a good idea. I'll update this.

PR #170, opened 2022-09-26
poc multitron

andreas-unleash:
This PR attempts to solve the misconfiguration of the unleash client
## About the changes

The main change is that the UnleashClient will only be available through the UnleashClientFactory.
The factory will hold instances of unleash client, so that whenever an attempt is made to recreate an existing client (one with the same config params `appName` and `unleashApi` ), instead of initialising a new instance, you will get the existing one - avoiding all the initial requests that are necessary for the client.

Closes # .

### DIFF
--- a/src/main/java/io/getunleash/DefaultUnleash.java
+++ b/src/main/java/io/getunleash/DefaultUnleash.java
@@ -41,10 +41,15 @@ public class DefaultUnleash implements Unleash {
     private final EventDispatcher eventDispatcher;
     private final UnleashConfig config;
 
-    private static FeatureRepository defaultToggleRepository(UnleashConfig unleashConfig) {
+    protected static FeatureRepository defaultToggleRepository(UnleashConfig unleashConfig) {
         return new FeatureRepository(unleashConfig);
     }
 
+    /**
+     * @deprecated This constructor will be removed in version 7.0.0
+     *     <p>Use {@link UnleashClientFactory#getInstance()}#getClient(UnleashConfig, Strategy)}
+     *     instead.
+     */
     public DefaultUnleash(UnleashConfig unleashConfig, Strategy... strategies) {
         this(unleashConfig, defaultToggleRepository(unleashConfig), strategies);
     }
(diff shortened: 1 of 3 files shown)


src/main/java/io/getunleash/UnleashClientFactory.java
```diff
@@ -0,0 +1,40 @@
...
+public class UnleashClientFactory {
+
+    private static UnleashClientFactory INSTANCE;
+
+    private static Map<String, DefaultUnleash> unleashClients = new HashMap<>();
+
+    private UnleashClientFactory() {}
+
+    public static synchronized UnleashClientFactory getInstance() {
+        if (INSTANCE == null) {
```
sjaanus, line 18: Is there a reason creating a instance of factory instead of using static methods and never initiating?

src/main/java/io/getunleash/UnleashClientFactory.java
```diff
@@ -0,0 +1,40 @@
...
+            UnleashConfig config, FeatureRepository featureRepository, Strategy... strategies) {
+        String key = config.getUnleashAPI() + config.getAppName();
+        if (unleashClients.containsKey(key)) {
+            System.out.println("Returning existing client");
+            return unleashClients.get(key);
+        }
+        DefaultUnleash unleash = new DefaultUnleash(config, featureRepository, strategies);
+        unleashClients.put(key, unleash);
+        return unleash;
+    }
```
chriswk, line 39: What do you do if the user instantiates clients with varying lists of custom strategies?

chriswk (commented):
This looks like a good starting point. Some concerns.

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -70,6 +75,7 @@ public DefaultUnleash(
             UnleashContextProvider contextProvider,
             EventDispatcher eventDispatcher,
             UnleashMetricService metricService) {
+        System.out.println("Creating new instance of unleash");
```
chriswk, line 78: Please don't keep this line. :P

PR #177, opened 2022-10-17
chorus: using more semantic java interface

### LINKED ISSUE #176: Why use a `BiFunction` when the desired interface was a `BiPredicate`?

In the `Unleash` interface, there are some methods with a `BiFunction<String, UnleashContext, Boolean>` argument. But the availables implementations for `Unleash` (`DefaultUnleash`, `FakeUnleash`) both have no safe measures against this function returning a `null`. Bot will yield a `NullPointerException`:

[`DefaultUnleash.java#L123-L127`](https://github.com/Unleash/unleash-client-java/blob/8525a8f56e2198bebcd893e169ae92b556cc8b07/src/main/java/io/getunleash/DefaultUnleash.java#L123-L127)

```java
        boolean enabled;
        UnleashContext enhancedContext = context.applyStaticFields(config);

        if (featureToggle == null) {
            enabled = fallbackAction.apply(toggleName, enhancedContext);
```

[`FakeUnleash.java#L43-L46`](https://github.com/Unleash/unleash-client-java/blob/8525a8f56e2198bebcd893e169ae92b556cc8b07/src/main/java/io/getunleash/FakeUnleash.java#L43-L46):

```java
    public boolean isEnabled(
            String toggleName, BiFunction<String, UnleashContext, Boolean> fallbackAction) {
        if (!features.containsKey(toggleName)) {
            return fallbackAction.apply(toggleName, UnleashContext.builder().build());
```

Also, it is more semantic to use `Predicate<T>` or similar instead of `Function<T, Boolean>` or similar.

### Background

Using `Predicate<T>` instead of `Function<T, Boolean>` is a lot more semantic, and also avoids (some) random null pointer exceptions that the java type system allows.

If the null pointer exception happens for some reason, it will be a violation of contract of the API consumer, not of the API provider, as it is now.

### Solution suggestions

Change the API to accept a `BiPredicate` instead of a `BiFunction`.

It is wildly expected that it shall be ok with the use of lambdas. It will break compatibility if someone passes directly an instance of `BiFunction<String, UnleashContext, Boolean>`, as the tests within here have broken.

Trying to accept both `BiPredicate<String, UnleashContext>` and `BiFunction<String, UnleashContext, Boolean>` will cause the compiler to get confused, as both will satisfy the lambda signature. For example, in `DefaultUnleash` will yield a compiler error:

```java
    @Override
    public boolean isEnabled(
            final String toggleName, final UnleashContext context, final boolean defaultSetting) {
        return isEnabled(
                toggleName,
                context,
                (n, c) -> defaultSetting);
    }
```

ambiguous method call, both functional interfaces match

This will require a manual action of the maintainer to specify that it is a `BiPredicate` that it is wanted:

```java
    @Override
    public boolean isEnabled(
            final String toggleName, final UnleashContext context, final boolean defaultSetting) {
        return isEnabled(
                toggleName,
                context,
                (BiPredicate<String, UnleashContext>) (n, c) -> defaultSetting);
    }
```

jeffque:
using more semantic java functional interface `BiPredicate` instead of `BiFunction`.

## About the changes

Has potential to break compatibility with previous versions (see [Discussion points](#Discussion-points)), but makes the code more semantic.

Closes #176.

### Important files

- src/main/java/io/getunleash/Unleash.java
- src/test/java/io/getunleash/UnleashTest.java

## Discussion points

This might break some API consumers if they are using directly the interface `BiFunction` instead of a lambda.

### DIFF
--- a/src/main/java/io/getunleash/DefaultUnleash.java
+++ b/src/main/java/io/getunleash/DefaultUnleash.java
@@ -15,7 +15,7 @@ import io.getunleash.util.ConstraintMerger;
 import io.getunleash.util.UnleashConfig;
 import io.getunleash.variant.VariantUtil;
 import java.util.*;
-import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -99,7 +99,7 @@ public class DefaultUnleash implements Unleash {
     @Override
     public boolean isEnabled(
             final String toggleName,
-            final BiFunction<String, UnleashContext, Boolean> fallbackAction) {
+            final BiPredicate<String, UnleashContext> fallbackAction) {
         return isEnabled(toggleName, contextProvider.getContext(), fallbackAction);
     }
 
@@ -107,7 +107,7 @@ public class DefaultUnleash implements Unleash {
     public boolean isEnabled(
             String toggleName,
             UnleashContext context,
-            BiFunction<String, UnleashContext, Boolean> fallbackAction) {
+            BiPredicate<String, UnleashContext> fallbackAction) {
         boolean enabled = checkEnabled(toggleName, context, fallbackAction);
         count(toggleName, enabled);
         eventDispatcher.dispatch(new ToggleEvaluated(toggleName, enabled));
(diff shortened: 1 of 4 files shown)


chriswk:
Hi @jeffque , sorry for the delayed answer. Thanks for this, it's a fair question. I think the honest answer to the question for why BiFunction rather than BiPredicate is that I wasn't aware that we had a BiPredicate. I knew of Predicate, Function and BiFunction, so should've deduced that there was a BiPredicate as well.

This is a breaking change, so will require a new major release. I still feel like it's the correct thing to do. Let me check with the team to verify if there's anything we really want to get in before a major, and I'll merge this.

Again, thank you!

chriswk (approved):
Much appreciated, and looks good

PR #179, opened 2022-11-15
fix: app name with slash or backslash

### LINKED ISSUE #175: Sanitize appName for backupFile
We're evaluating feature flags on GitLab. The first tests are promising.
Sadly we ran into issues with backup files on Windows.
GitLab enforces to set application name to their "environment". This ist eine for some correlation.
See also: https://docs.gitlab.com/ee/operations/feature_flags.html#get-access-credentials

Our environment names are based on branch names. Thus we've environments like follow:
- feature/foo/bare
- fix/baz-123456

On Windows this yield into invalid path names.
See: 
https://github.com/Unleash/unleash-client-java/blob/8525a8f56e2198bebcd893e169ae92b556cc8b07/src/main/java/io/getunleash/util/UnleashConfig.java#L605

I recommend to sanitize appName for backup files.
What do you think?

ajdergute:
## About the changes

This PR removes slashes and backslashes from appName, as they cause issues on windows/linux. Both are replaced by "-".
Additionally a check is added to prevent a double (back)slash in some cases when reading `java.io.tmpdir`.

Closes #175

### DIFF
--- a/src/main/java/io/getunleash/util/UnleashConfig.java
+++ b/src/main/java/io/getunleash/util/UnleashConfig.java
@@ -602,8 +602,18 @@ public class UnleashConfig {
             if (backupFile != null) {
                 return backupFile;
             } else {
-                String fileName = "unleash-" + appName + "-repo.json";
-                return System.getProperty("java.io.tmpdir") + File.separatorChar + fileName;
+                String fileName = "unleash-" + sanitizedAppName(appName) + "-repo.json";
+                String tmpDir = System.getProperty("java.io.tmpdir");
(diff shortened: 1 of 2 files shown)


chriswk (approved):
:+1: Thank you for this. Good point about being compatible with Windows platform as well.

ajdergute:
@chriswk, I'm not familiar with the process. I fixed tests. Please let me know if you need something different.
Otherwise could you plz approve this workflow.

chriswk:
Great Job ajder. I'll merge, and you'll be a contributor in our next release.

Thanks for your contribution. :)

PR #180, opened 2022-11-24
fix: getVariant counts enabled or not

chriswk:
As a change from earlier, this now increments yes/no as well as which variant was received in the MetricsBucket.

To reflect behaviour of Node SDK from https://github.com/Unleash/unleash-client-node/commit/38f35e658a34c510b4d01f8dd33866ff1119d96e onwards.


sighphyre (approved):
LGTM

PR #183, opened 2022-12-26
Validate behaviour with empty variants

gastonfournier:
## About the changes
This PR intends to test the behaviour when variants are empty. Related with https://github.com/Unleash/client-specification/pull/45

## Discussion points
What should be the right behaviour in this situation?


gastonfournier:
The behavior is already validated here: https://github.com/Unleash/unleash-client-java/blob/be22d603ae9a37cb0617e6ac3397e823c244f9fa/src/test/java/io/getunleash/variant/VariantUtilTest.java#L41-L49 and it's the correct one

Issue #181, opened 2022-12-16
DefaultUnleash.isEnabled(toggleName, context) silently dropping context


hnicke:
### Describe the bug

The `Unleash` Interface provides a function `isEnabled(toggleName,  context)` with a default implementation.
In the default implementation, the context is silently dropped.
This is expected behavior i guess.

However, he `DefaultUnleash` implementation is missing an override for this function.
This leads to the behavior where `DefaultUnleash.isEnabled(toggleName, context)` silenlty drops the context.
Note that `DefaultUnleash.isEnabled(toggleName, context, defaultValue)` does not drop the context.

### SDK information (language and version)

java client sdk 7.0.0

pransh15:
Hey @hnicke 👋 
Thanks for submitting this, someone from the team will get back to you on this soon. 😄

gastonfournier:
Hi @hnicke the `DefaultUnleash` implementation seems to be working as expected (check #184 where I explained why is that). The interface is confusing nevertheless, so I'll bring this to @chriswk as an enhancement.

hnicke:
Thank you for looking into this.
My bad - you are right!
For some odd reason it didn't work for me last week.
I can confirm it is working as intended.
Sorry for the inconvenience.

gastonfournier:
No problem, still it is something that might be worth looking at!

PR #184, opened 2022-12-26
Test isEnabled with context

gastonfournier:
## About the changes
This is just a test to validate #181 

This validates that the context is not being dropped. 

Looking into the details:
1. isEnabled with toggle name and context calls isEnabled with toggle name, context and `false` as defaultSetting: https://github.com/Unleash/unleash-client-java/blob/3e4b5250de49b801cd054dae92480878403e5219/src/main/java/io/getunleash/Unleash.java#L11-L13
2. Certainly the default implementation is dropping the context: https://github.com/Unleash/unleash-client-java/blob/3e4b5250de49b801cd054dae92480878403e5219/src/main/java/io/getunleash/Unleash.java#L15-L17
3. But that default method is not invoked because the default client overrides the default implementation: https://github.com/Unleash/unleash-client-java/blob/3e4b5250de49b801cd054dae92480878403e5219/src/main/java/io/getunleash/DefaultUnleash.java#L93-L97

## Discussion points
1. The default methods in the interface seem to be shortcuts for testing (FakeUnleash) but they don't provide the proper implementation for a production system that may want to implement `Unleash` interface (they should remember to implement that method). For being error-prone I consider we should not have a default implementation for `Unleash` interface, but that's a breaking change and we should discuss it
2. I get also confused by another method dropping the `fallbackAction` and it might be better to have a method receiving all the parameters and this should be the only one you should implement: `isEnabled(toggleName, context, fallbackAction, defaultSetting);` but I need to dig deeper on the meaning of those parameters and if it makes sense to have both. 
 
I'll talk about this with @chriswk who may have more context

### DIFF
--- a/src/test/java/io/getunleash/DefaultUnleashTest.java
+++ b/src/test/java/io/getunleash/DefaultUnleashTest.java
@@ -1,6 +1,7 @@
 package io.getunleash;
 
 import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.*;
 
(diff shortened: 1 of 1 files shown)


gastonfournier:
Closing as this use case should be already covered by client specification tests, in particular https://github.com/Unleash/client-specification/blob/main/specifications/02-user-with-id-strategy.json

PR #185, opened 2022-12-26
feat: proposal to change default methods in Unleash interface

gastonfournier:
## About the changes
Proposal based on #181:
1. Adds default empty context to avoid having to deal with nulls
2. Transforms `false` into a default fallback action
3. Leaving `boolean isEnabled(String toggleName, UnleashContext context, BiPredicate<String, UnleashContext> fallbackAction)` as the only method that needs to be implemented


chriswk (approved):
LGTM

PR #186, opened 2023-01-02
Task/detect multiple instantiations

chriswk:
## What
This PR introduces a `ConcurrentHashMap<String, LongAdder>` for keeping track of the number of instantiations per client identifier (set to be apiKey, appName, instanceId). If the client instantiates Unleash with the same identifier twice a log ERROR line will be printed. This should allow us to more easily help customers using this SDK that have problems with their applications connecting way too often.

## Discussion
It will be worth it to replace the DefaultUnleash constructor with a builder that keeps track of configs seen and static fields with already instantiated clients to hopefully avoid the excessive instantiation altogether, but that would require making the DefaultUnleash constructors private and would constitute a major release. We should plan for including the builder + constructor access change for the next major release. Which should come when we implement dependent feature toggles.

## Interesting files
DefaultUnleash.java
DefaultUnleashTest.java


gastonfournier (approved):

sighphyre (changes requested):
I think this looks good, could use some docs/README updates though before we merge it (this is the requested change). I've also left a comment on how I think this should work but I'm open to being told to go away on that

gastonfournier:
> I think this looks good, could use some docs/README updates though before we merge it (this is the requested change).

I have added this note: https://github.com/Unleash/unleash-client-java/pull/186/commits/ee786c331d1d2ad8f896a9c225567f487e6b8183

This flag could be a configuration value and then be in the table where we describe all configuration values. But when we talked about this with @chriswk we concluded this is not something we want to support in the long run and will likely be a breaking change for the next major SDK release (and this constructor will disappear), but we do want to enable the consumers of this SDK to opt-in or be early adopters of the feature.

sighphyre (approved):
LGTM

PR #189, opened 2023-01-05
docs: merge Java doc with docs.getunleash content

thomasheartman:
## What

This change attempts to integrate all the information that currently
exists in the [docs.getunleash Java
doc](https://docs.getunleash.io/reference/sdks/java) into the readme.

It also updates some of the links to the doc from the old format to
the new format.

## Why

Having multiple versions of the same documentation can be confusing
and cause drift. We're moving to automatically generating the SDK
content on docs.getunleash now (from GitHub readmes), so it's
important to ensure that we don't lose any information in the process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ You need to add the Unleash SDK as a dependency for your project. Here's how you
 
 **⚠️ Important:** In almost every case, you only want a **single, shared instance of the `Unleash`  class (a *singleton*)** in your application . You would typically use a dependency injection framework (such as Spring or Guice) to inject it where you need it. Having multiple instances of the client in your application could lead to inconsistencies and performance degradation.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,40 +24,58 @@ You will require unleash on your class path, pop it in to your pom:
...
 UnleashConfig config = UnleashConfig.builder()
-            .appName("java-test")
-            .instanceId("instance x")
-            .unleashAPI("http://unleash.herokuapp.com/api/")
-            .apiKey("API token")
-            .build();
+        .appName("my.java-app")
+        .instanceId("your-instance-1")
+        .unleashAPI("<unleash-api-url>")
+        .customHttpHeader("Authorization", "<client-api-token>")
```
chriswk, line 55: Still recommend using `apiKey("<client-api-token")` instead of `customHttpHeader("")`

thomasheartman, line 55: Oh, for sure! That's much nicer. I'll go ahead and update the uses.

PR #195, opened 2023-03-29
feat: Adds validate method to UnleashConfig

chriswk:
User request to have a way to validate that configuration was correct that can actually interrupt controlflow. The Eventmodel wasn't suited for this, since the event dispatcher dispatches on a different thread to the callee. This adds a way to have the exception thrown on the main caller thread.

### DIFF
--- a/src/main/java/io/getunleash/util/UnleashConfig.java
+++ b/src/main/java/io/getunleash/util/UnleashConfig.java
@@ -5,6 +5,7 @@ import static io.getunleash.DefaultUnleash.UNKNOWN_STRATEGY;
 import io.getunleash.CustomHttpHeadersProvider;
 import io.getunleash.DefaultCustomHttpHeadersProviderImpl;
 import io.getunleash.UnleashContextProvider;
+import io.getunleash.UnleashException;
 import io.getunleash.event.NoOpSubscriber;
 import io.getunleash.event.UnleashSubscriber;
 import io.getunleash.lang.Nullable;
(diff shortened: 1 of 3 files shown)


src/main/java/io/getunleash/util/UnleashConfig.java
```diff
@@ -185,6 +186,16 @@ private void enableProxyAuthentication() {
...
     }
 
+    /**
+     * Fetches client features using this configuration.
+     * If the method does not throw, the configuration was able to fetch features and as such is assumed to be correct.
+     * Use upon initial building of an UnleashClient to simplify debugging and interrupting your application if client is incorrectly configured.
+     * @throws UnleashException Thrown if something went wrong when trying to fetch features. Typical errors are incorrect URLs or apiKeys
+     */
+    public void validate() throws UnleashException {
+        this.getUnleashFeatureFetcherFactory().apply(this).fetchFeatures();
```
gastonfournier, line 196: I'm not sure if this will solve all the issues, maybe we should also assert that the response status is not `UNAVAILABLE`: https://github.com/Unleash/unleash-client-java/blob/e5c3e34d6c47bb1c233ae09db293f17583b33e51/src/main/java/io/getunleash/repository/OkHttpFeatureFetcher.java#L93-L94

ivarconr:
I wonder, have you considered if using the `synchronousFetchOnInitialisation` option would be enough to trigger the validation on inialization?

src/main/java/io/getunleash/util/UnleashConfig.java
ivarconr: I agree with @gastonfournier we should make sure to validate that the response is 200. 

I wonder why you would do the validation as part of the configuration object instead of having a similar function on the client itself, then we could make use of the actual response as well as the initial toggle values.

gastonfournier:
> I wonder, have you considered if using the `synchronousFetchOnInitialisation` option would be enough to trigger the validation on inialization?

I'm testing that out, and added some tests. It seems it wasn't working as expected, and I'm trying to fix that

chriswk:
After discussing internally, and realising that this new method only uses already public methods, so we will not expose this as a separate function. Closing this PR

PR #196, opened 2023-03-30
fix: synchronous fetch on initialization

gastonfournier:
## About the changes
When `synchronousFetchOnInitialization` is enabled, we should fail synchronously if the initial fetch fails.

Currently, we're swallowing the exception

There are a few sonatype-lift recommendations which seem reasonable, but I'm not sure I'd like to double check with @chriswk


src/test/java/io/getunleash/example/UnleashUsageTest.java
```diff
@@ -15,7 +15,7 @@ public void wire() {
                 new UnleashConfig.Builder()
                         .appName("test")
                         .instanceId("my-hostname:6517")
-                        .synchronousFetchOnInitialisation(true)
+                        .synchronousFetchOnInitialisation(false)
```
gastonfournier, line 18: There's no localhost:4242 set up in this test, so this test would fail if we fetch synchronously

ivarconr (approved):

src/test/java/io/getunleash/repository/FeatureToggleRepositoryTest.java
```diff
@@ -156,6 +157,27 @@ public void should_perform_synchronous_fetch_on_initialisation() {
         verify(toggleFetcher, times(1)).fetchToggles();
     }
 
+    @Test
+    public void should_perform_synchronous_fetch_on_initialisation_and_fail_if_invalid_upstream() {
```
ivarconr, line 161: Can we add a test for the opposite as well, that we should not throw if` `synchronousFetchOnInitialisation=false`

gastonfournier: We have many tests with this configuration, but I believe we have no test validating that after the initial failure the next interval we're going to try again. Let me check and add that

gastonfournier: Added here: [fef2827](https://github.com/Unleash/unleash-client-java/pull/196/commits/fef2827de4bad7290ba77c7ee06f660b12916987) and checked with @chriswk 
I'll proceed with the merge

PR #198, opened 2023-04-05
fix: Update random to 0-10000 instead of 0-100

chriswk:
After thorough testing with Go, Java and Javascript the last few days, we've found that using 0-100 didn't give enough entropy to the first bytes of the hash bucket, causing odd distributions. This updates to using 10000 as the max range to increase entropy and reflect what we do in node (since January) and in go (since April).


sighphyre (approved):
LGTM!

PR #201, opened 2023-06-14
feat: expose impression data property

sighphyre:
This exposes the impression data property on the FeatureToggle class through a getter - `hasImpressionData()` this will default to false if not present

### DIFF
--- a/src/test/java/io/getunleash/repository/JsonFeatureToggleParserTest.java
+++ b/src/test/java/io/getunleash/repository/JsonFeatureToggleParserTest.java
@@ -118,7 +118,7 @@ public class JsonFeatureToggleParserTest {
     }
 
     @Test
-    public void should_should_deserialize_impression_data() throws IOException {
+    public void should_deserialize_impression_data() throws IOException {
         Reader content = getFileReader("/unleash-repo-v2-with-impression-data.json");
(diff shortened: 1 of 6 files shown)


sighphyre:
Hmm, okay, my built in formatter has made changes. Applying the maven task doesn't make changes here, happy to undo the changes by hand if folks object

src/test/java/io/getunleash/repository/JsonFeatureToggleParserTest.java
```diff
@@ -117,6 +117,21 @@ public void should_deserialize_list_of_toggles_with_variants() throws IOExceptio
                 .isNull();
     }
 
+    @Test
+    public void should_should_deserialize_impression_data() throws IOException {
```
chriswk, line 121:
```suggestion
    public void should__deserialize_impression_data() throws IOException {
```

chriswk (approved):
LGTM

src/test/java/io/getunleash/repository/JsonFeatureToggleParserTest.java
sighphyre: Yeah, it really really should. It should should

Jokes aside, nice catch! Thanks

PR #204, opened 2023-07-19
fix: bump unleash-client-java version in examples

andreas-unleash:
Bump unleash client version to 8.2.0 in both examples
## About the changes

Closes #


sighphyre (approved):
Nice!

Issue #206, opened 2023-08-07
SDK exposes Api key in logs


chriswk:
### Describe the bug

When instantiating more than one client in the same process, Unleash warns you that this is not a recommended way to do so. When it warns you the ERROR log message contains the API key it has seen more than one client instance for.

### Steps to reproduce the bug

Instantiate two UnleashClients in the same process from the same Unleash Config.
Once the second client is instantiated, observe that the error message contains the api key

### Expected behavior

I'd still like to see the ERROR message, but with a redacted API key.

### Subscription type

None

### Hosting type

None

### SDK information (language and version)

unleash-client-java:8.2.0

PR #207, opened 2023-08-07
fix: Use sha256 digest instead of exposing api key

chriswk:
As mentioned in #206 - The way we did the identifier for a client inadvertently also exposes api keys in the logs when logging that you have more than one client using the same configuration. This fix uses the SHA-256 hash of the three elements (apiKey, appName and instanceId) together and then uses the hex representation of  that as the client identifier instead.

In addition this updates the log message to tell you that you have n clients for appName:instanceId combination.


sighphyre (approved):
LGTM!

PR #209, opened 2023-08-21
task: Added workflow for releasing new versions from GHA

chriswk:
If this works, it would make the workflow the same as we have for Unleash server. 
Trigger the workflow, type the desired version to release and wait.


andreas-unleash (approved):

andreas-unleash:
Nice

PR #211, opened 2023-09-04
fix: when using multiple strategies some of them might not be evaluated

### LINKED ISSUE #210: Strategy check return only the first enabled
### Describe the bug

In the [commit https://github.com/Unleash/unleash-client-java/commit/78a0d33358bb375d6abe8d08de60f704a12961a6](feat: strategy variants) has been introduced a bug.
When retrieving the toggle enabled strategies it is returned only the first. In case of multiple strategy the the toggle will result disabled even when the context is valid.

See into class `src/main/java/io/getunleash/DefaultUnleash.java`
```
        } else {
            enabledStrategy =
                    featureToggle.getStrategies().stream()
                            .filter(
                                    strategy -> {
                                        Strategy configuredStrategy =
                                                getStrategy(strategy.getName());
	@@ -181,13 +184,33 @@ private boolean checkEnabled(
                                        }

                                        return configuredStrategy.isEnabled(
                                                strategy.getParameters(), context);
                                    })
                            .findFirst();
```

### Steps to reproduce the bug

1. Create a toggle with more then one strategy
2. Check if the toggle is enabled having as context valued that meet second or further strategy in the list.
3. Should expect to return enabled true but is false

### Expected behavior

When calling the method `isEnabled` with context that meet one of the strategy should expect `true` as response

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### Unleash version

8.3.0 ( introduced in this commit-78a0d33358bb375d6abe8d08de60f704a12961a6)

### Subscription type

Enterprise

### Hosting type

Hosted by Unleash

### SDK information (language and version)

java

gastonfournier:
## About the changes
The code selecting an activation strategy was exiting on a first enable match but the strategy being enabled does not mean the feature will be enabled. We need to iterate and evaluate each one of the strategies before returning.

Closes #210

### DIFF
--- a/src/main/java/io/getunleash/DefaultUnleash.java
+++ b/src/main/java/io/getunleash/DefaultUnleash.java
@@ -202,7 +202,7 @@ public class DefaultUnleash implements Unleash {
                 }
             }
         }
-        return new FeatureEvaluationResult(enabled, null);
+        return new FeatureEvaluationResult(enabled, defaultVariant);
     }
 
     private void checkIfToggleMatchesNamePrefix(String toggleName) {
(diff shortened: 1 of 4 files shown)


sighphyre (approved):
G

PR #213, opened 2023-09-05
feat: make sure gpg entry expects passphrase in environment variable

chriswk:
As we saw when we tried to release 8.3.1, we had an `inappropriate ioctl for device`. This PR adds a pinentry rule to loopback, so that we can pass in the passphrase as environment variable. In addition it switches the scm URLs to point to https URLs so our maven process is able to update the github repository with tags. 

Tested in https://github.com/chriswk/javarelease, we know the publishing to maven central repo works, but we need to know that the git tagging works as well.


sighphyre (approved):
LGTM

PR #217, opened 2023-10-03
docs: make synchronousFetchOnInitialisation in option more clear

alexander-svendsen:
## About the changes
just to make it more clear in case someone skipped section about Asynchronous/synchronous initialization


ivarconr (approved):
Feels a bit like "butter on bacon" given the parameter is called "synchronousFetchOnInitialisation" . But as long as it makes it clearer.

PR #218, opened 2023-10-12
feat: Handle dependent feature toggles

chriswk:
- A new feature added in beta to Unleash 5.6. With dependent feature toggle you can configure parent-child relationships. This change makes the SDK evaluate according to the client specifications defined for this.

### DIFF
--- a/src/main/java/io/getunleash/DefaultUnleash.java
+++ b/src/main/java/io/getunleash/DefaultUnleash.java
@@ -136,17 +136,17 @@ public class DefaultUnleash implements Unleash {
 
     @Override
     public boolean isEnabled(
-        String toggleName,
-        UnleashContext context,
-        BiPredicate<String, UnleashContext> fallbackAction) {
+            String toggleName,
+            UnleashContext context,
+            BiPredicate<String, UnleashContext> fallbackAction) {
         return isEnabled(toggleName, context, fallbackAction, false);
     }
 
     public boolean isEnabled(
-        String toggleName,
-        UnleashContext context,
-        BiPredicate<String, UnleashContext> fallbackAction,
-        boolean isParent) {
+            String toggleName,
+            UnleashContext context,
+            BiPredicate<String, UnleashContext> fallbackAction,
+            boolean isParent) {
         FeatureEvaluationResult result =
                 getFeatureEvaluationResult(toggleName, context, fallbackAction, null);
         if (!isParent) {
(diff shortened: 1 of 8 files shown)


src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -168,40 +169,89 @@ private FeatureEvaluationResult getFeatureEvaluationResult(
...
-            for (ActivationStrategy strategy : featureToggle.getStrategies()) {
-                Strategy configuredStrategy = getStrategy(strategy.getName());
-                if (configuredStrategy == UNKNOWN_STRATEGY) {
-                    LOGGER.warn(
-                            "Unable to find matching strategy for toggle:{} strategy:{}",
-                            toggleName,
-                            strategy.getName());
-                }
+            // Dependent toggles, no point in evaluating child strategies if our dependencies are
+            // not enabled
```
kwasniew, line 177: not enabled -> not satisfied. satisfied may mean disabled or enabled depending on how we configure our dependencies

src/test/java/io/getunleash/DefaultUnleashTest.java
```diff
@@ -275,4 +272,18 @@ public void client_identifier_handles_api_key_being_null() {
         assertThat(id)
                 .isEqualTo("f83eb743f4c8dc41294aafb96f454763e5a90b96db8b7040ddc505d636bdb243");
     }
+
+    /*    @Test
```
kwasniew, line 276: what's up with this test?

kwasniew (approved):
Very nice. A few minor comments inside. 2 things I'd test since they are not part of the client specification:
* metrics are not called on dependent features
* impression events are called on dependent features for easier debugging

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -168,40 +169,89 @@ private FeatureEvaluationResult getFeatureEvaluationResult(
...
-                    result.setVariant(variant);
-                    return result;
                 }
             }
             return new FeatureEvaluationResult(false, defaultVariant);
         }
     }
 
+    //
+    // Checks a feature's dependencies (if it has any
```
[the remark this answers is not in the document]
chriswk: yeah. thanks

src/test/java/io/getunleash/DefaultUnleashTest.java
chriswk: It isn't needed, so I'll remove it

andreas-unleash (approved):
LGTM

PR #219, opened 2023-10-24
docs: Updated docs to reference actually existing classes

chriswk:
@sjaanus pointed out that we were referencing classes he couldn't find in the source code. After looking through, I see that we were indeed incorrectly naming the classes used. This PR updates the two references referred to.


sjaanus (approved):
LGTM

PR #221, opened 2023-11-07
feat: Obey HTTP Status codes from upstream.

chriswk:
Previously we have treated any status code as equal. This PR changes that to allow 429 and 50x status codes to incrementally increase the interval between each time we call. And then gradually decrease once it starts succeeding.

### Discussion points

- a bit worried about the metrics here, because we drop the metric bucket on the floor if we don't get a 20x back from the server. We did that previously as well, so there's nothing new here. A failing call to sendMetrics would dump the metrics on the floor, since we new up a new bucket before sending the previous one.

### DIFF
--- a/src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
+++ b/src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
@@ -1,12 +1,11 @@
 package io.getunleash.metric;
 
+import io.getunleash.util.Throttler;
 import io.getunleash.util.UnleashConfig;
 import io.getunleash.util.UnleashScheduledExecutor;
-import java.net.HttpURLConnection;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicInteger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -20,9 +19,7 @@ public class UnleashMetricServiceImpl implements UnleashMetricService {
     // mutable
     private volatile MetricsBucket currentMetricsBucket;
 
-    private final int maxSkips;
-    private final AtomicInteger failures = new AtomicInteger();
-    private final AtomicInteger skips = new AtomicInteger();
+    private final Throttler throttler;
 
     public UnleashMetricServiceImpl(
             UnleashConfig unleashConfig, UnleashScheduledExecutor executor) {
@@ -37,8 +34,11 @@ public class UnleashMetricServiceImpl implements UnleashMetricService {
         this.started = LocalDateTime.now(ZoneId.of("UTC"));
         this.unleashConfig = unleashConfig;
         this.metricSender = metricSender;
-        this.maxSkips =
-                Integer.max(20, 300 / Integer.max((int) unleashConfig.getSendMetricsInterval(), 1));
+        this.throttler =
+                new Throttler(
+                        (int) unleashConfig.getSendMetricsInterval(),
+                        300,
+                        unleashConfig.getUnleashURLs().getClientMetricsURL());
         long metricsInterval = unleashConfig.getSendMetricsInterval();
         executor.setInterval(sendMetrics(), metricsInterval, metricsInterval);
     }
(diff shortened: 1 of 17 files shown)


src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
```diff
@@ -28,6 +38,7 @@ public UnleashMetricServiceImpl(
         this.started = LocalDateTime.now(ZoneId.of("UTC"));
         this.unleashConfig = unleashConfig;
         this.metricSender = metricSender;
+        this.maxInterval = Integer.max(20, 300 / Integer.max(Long.valueOf(unleashConfig.getSendMetricsInterval()).intValue(), 1));
```
chriswk, line 41: Happy to discuss this, our goal was that the max should equal 300 seconds at most. because 5 minute interval seems to be enough. I realize the metric sending could be bigger. Suggestions?

ivarconr:
> drop the metric bucket on the floor if we don't get a 20x back from the server.

I agree this is a concern but I think it needs to be solved in a separate PR. Other SDKs will preserve the metrics and send them on the next interval.

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
```diff
@@ -28,6 +37,14 @@ public UnleashMetricServiceImpl(
         this.started = LocalDateTime.now(ZoneId.of("UTC"));
         this.unleashConfig = unleashConfig;
         this.metricSender = metricSender;
+        this.maxInterval =
+                Integer.max(
+                        20,
+                        300
+                                / Integer.max(
+                                        Long.valueOf(unleashConfig.getSendMetricsInterval())
+                                                .intValue(),
```
gastonfournier, line 46: According to Long#intValue method:
```java
    /**
     * Returns the value of this {@code Long} as an {@code int} after
     * a narrowing primitive conversion.
     * @jls 5.1.3 Narrowing Primitive Conversion
     */
    public int intValue() {
        return (int)value;
    }
```

So this should be safe:
```suggestion
                                / Integer.max(
                                        (int) unleashConfig.getSendMetricsInterval(),
```

src/main/java/io/getunleash/repository/FeatureRepository.java
```diff
@@ -90,8 +122,13 @@ private void initCollections(UnleashScheduledExecutor executor) {
         }
     }
 
+    //
```
gastonfournier, line 125:
```suggestion

```

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
```diff
@@ -28,6 +37,14 @@ public UnleashMetricServiceImpl(
         this.started = LocalDateTime.now(ZoneId.of("UTC"));
         this.unleashConfig = unleashConfig;
         this.metricSender = metricSender;
+        this.maxInterval =
+                Integer.max(
+                        20,
+                        300
```
gastonfournier, line 43: Is maxInterval in seconds? What are these magic numbers for? 

I'm struggling to see why the largest sendMetricsInterval is, the smaller maxInterval becomes. Or it's too late for math atm.

Ok, on second thought, is it the number of times we want to skip/avoid making an HTTP call? If that's the case I'd suggest renaming the variable to `skip` and `maxSkips` maybe, as Interval has a different meaning in the java.time API which might mislead readers of this code

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
```diff
@@ -51,11 +68,62 @@ public void countVariant(String toggleName, String variantName) {
...
+        } else if (responseCode == HttpURLConnection.HTTP_UNAUTHORIZED
+                || responseCode == HttpURLConnection.HTTP_FORBIDDEN) {
+            failures.incrementAndGet();
+            interval.set(maxInterval);
+            LOGGER.error(
+                    "Client was not authorized to post metrics to the Unleash API at {}. Backing off to {} times our poll interval to avoid overloading server",
+                    unleashConfig.getUnleashURLs().getClientMetricsURL(),
+                    maxInterval);
+        } else if (responseCode >= 500) {
+            interval.set(Math.min(failures.incrementAndGet(), maxInterval));
```
gastonfournier, line 114: I'm thinking these modifications of the interval can be added to an internal method like `this.increaseInterval()` or `this.skipMore()` (depending on the name). I believe that can improve the maintainability of the code.

src/test/java/io/getunleash/event/SubscriberTest.java
```diff
@@ -44,19 +45,29 @@ void setup() {
...
+        assertThat(testSubscriber.togglesFetchedCounter)
+                .isEqualTo(
+                        2); // Server successfully returns, we call synchronous fetch and schedule
+        // once, so 2 calls.
+        assertThat(testSubscriber.status).isEqualTo(CHANGED);
         assertThat(testSubscriber.toggleEvaluatedCounter).isEqualTo(3);
         assertThat(testSubscriber.toggleName).isEqualTo("myFeature");
         assertThat(testSubscriber.toggleEnabled).isFalse();
-        assertThat(testSubscriber.errors).hasSize(2);
+        assertThat(testSubscriber.errors).isEmpty();
```
gastonfournier, line 70: Yes, this looks better :+1:

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
chriswk: Agreed, will extract method

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
chriswk: Good point, I'll rename. Yes, it is the number of times we're going to skip/do nothing.

gastonfournier (approved):
LGTM! Just a small comment about externalizing the throttling behavior into an instance so we don't repeat the code, but other than that this looks neat!

chriswk:
> LGTM! Just a small comment about externalizing the throttling behavior into an instance so we don't repeat the code, but other than that this looks neat!

Handled with adding the throttler class.

PR #222, opened 2023-11-08
fix: prevent failure count going below zero

chriswk:
While testing this, we found that successful invocation reduced failure count below 0, this caused the backoff to not work as intended


gardleopard (approved):
Great! We testet this against a sandbox instance as well.

PR #225, opened 2023-11-21
chore: allow to fail if unavailable

gastonfournier:
## About the changes
On synchronous initialization, we were not failing in case of an invalid response code but still marking the client as ready. With this change, we're making sure we have a valid features collection before marking the client as ready and we're not going to throttle in case of asynchronous initialization.

### DIFF
--- a/src/main/java/io/getunleash/repository/FeatureRepository.java
+++ b/src/main/java/io/getunleash/repository/FeatureRepository.java
@@ -110,10 +110,6 @@ public class FeatureRepository implements IFeatureRepository {
         }
     }
 
-    private Integer calculateMaxSkips(int fetchTogglesInterval) {
-        return Integer.max(20, 300 / Integer.max(fetchTogglesInterval, 1));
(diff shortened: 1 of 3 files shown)


chriswk (approved):
Should we make exception handler configurable here?

chriswk:
I see our tests that test that we increment failure count when we fail to connect is failing.

src/main/java/io/getunleash/repository/FeatureRepository.java
```diff
@@ -108,11 +110,7 @@ private void initCollections(UnleashScheduledExecutor executor) {
         }
     }
 
-    private Integer calculateMaxSkips(int fetchTogglesInterval) {
-        return Integer.max(20, 300 / Integer.max(fetchTogglesInterval, 1));
-    }
```
gastonfournier, line 113: This was unused

sighphyre (approved):
G

PR #229, opened 2023-11-29
chore: remove 3 hostname tests that are no longer relevant

sighphyre:
This removes 3 of the hostname strategy tests that are unlikely to ever work correctly here and probably no longer belong. This logic is now owned by Yggdrasil and has relevant tests covering the behaviour that these 3 tests were previously testing.

Usually not a fan of removing tests to make things work but:

1) These tests now effectively test the underlying library
2) They require setting the hostname env variable. Doing this through Java only the affects JVM, not the Rust code
3) At least one test remains that proves the behaviour works across the FFI boundary (gets the hostname in Java without modification, Rust gets the same hostname later)

### DIFF
--- a/src/test/java/io/getunleash/strategy/ApplicationHostnameStrategyTest.java
+++ b/src/test/java/io/getunleash/strategy/ApplicationHostnameStrategyTest.java
@@ -72,39 +72,6 @@ public class ApplicationHostnameStrategyTest {
     }
 
     @Test
-    public void should_be_enabled_for_hostName() {
-        String hostName = "my-super-host";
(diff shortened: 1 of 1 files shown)


chriswk (changes requested):
Please remove the so and jar files before committing this.

chriswk (approved):
As long as these binaries stay on branch only, I'm :+1: on removing this.

gastonfournier (approved):
Makes sense as the SDK is no longer reading the hostname, but Yggdrasil is doing so, these tests at most should be inside Yggdrasil to make sure it can handle such cases (I imagine the tests are there for a reason)

PR #240, opened 2024-05-06
fix: check that contextValue starts with

### LINKED ISSUE #238: String operator STR_STARTS_WITH is not evaluated correctly when case-sensative is disabled
### Describe the bug

When using the STR_STARTS_WITH operator as a constraint on any context field with case-sensitive de-activated, the constraint is always evaluates to false.

### Steps to reproduce the bug

Create an activation strategy for an enabled feature-toggle and update it with the following definition
```
curl --location --request PUT '<your-host>/api/admin/projects/<your-project>/features/test/environments/<your-environment>/strategies/<your-strategy-id>' \
    --header 'Authorization: INSERT_API_KEY' \
    --header 'Content-Type: application/json' \
    --data-raw '{
  "name": "flexibleRollout",
  "title": "",
  "constraints": [
    {
      "values": [
        "0"
      ],
      "inverted": false,
      "operator": "STR_STARTS_WITH",
      "contextName": "userId",
      "caseInsensitive": true
    }
  ],
  "parameters": {
    "rollout": "100",
    "stickiness": "default",
    "groupId": "test"
  },
  "variants": [
    {
      "stickiness": "default",
      "name": "a",
      "weight": 1000,
      "payload": {
        "type": "string",
        "value": "a"
      },
      "weightType": "variable"
    }
  ],
  "segments": [],
  "disabled": false
}'
```
Now evaluate the feature toggle you updated the strategy for with any user-id that starts with "0".
The result is "disabled".

### Expected behavior

The evaluation result should return the variation with the name "a".

### Logs, error output, etc.

_No response_

### Screenshots

_No response_

### Additional context

When looking into the code the error is easy to [spot](https://github.com/Unleash/unleash-client-java/blob/b2ac9aee6aecbd3e968ed5c0e51bb5fca1c60452/src/main/java/io/getunleash/strategy/constraints/StringConstraintOperator.java#L66 )
```java
return contextValue
                .map(
                        c ->
                                values.stream()
                                        .anyMatch(
                                                v -> {
                                                    if (caseInsensitive) {
                                                        return v.toLowerCase(comparisonLocale) // <-- v.startsWith should be c.startsWith
                                                                .startsWith(
                                                                        c.toLowerCase(
                                                                                comparisonLocale));
                                                    } else {
                                                        return c.startsWith(v); // <-- this is correct
                                                    }
                                                }))
                .orElse(false);
```
The comparison is the wrong way around as can be seen in the else-branch or any other method in this class, where it's done correctly.

### Unleash version

_No response_

### Subscription type

None

### Hosting type

None

### SDK information (language and version)

unleash-client-java:9.2.0

chriswk:
There had been an inversion of variable usage for one of our cases in the matcher. This PR makes sure to compare contextValue to see if it starts with the requested value in the constraint, instead of the other way around.

fixes #238

### DIFF
--- a/src/main/java/io/getunleash/strategy/constraints/StringConstraintOperator.java
+++ b/src/main/java/io/getunleash/strategy/constraints/StringConstraintOperator.java
@@ -58,17 +58,17 @@ public class StringConstraintOperator implements ConstraintOperator {
             List<String> values, Optional<String> contextValue, boolean caseInsensitive) {
         return contextValue
                 .map(
-                        c ->
+                    actualContextValue ->
                                 values.stream()
                                         .anyMatch(
-                                                v -> {
+                                            value -> {
                                                     if (caseInsensitive) {
-                                                        return v.toLowerCase(comparisonLocale)
+                                                        return actualContextValue.toLowerCase(comparisonLocale)
                                                                 .startsWith(
-                                                                        c.toLowerCase(
+                                                                        value.toLowerCase(
                                                                                 comparisonLocale));
                                                     } else {
-                                                        return c.startsWith(v);
+                                                        return actualContextValue.startsWith(value);
                                                     }
                                                 }))
                 .orElse(false);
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
Nice fix, silly bug

PR #242, opened 2024-05-08
chore: Use client spec 5.1.5 instead of 5.0.2

chriswk:
To make sure the SDK is up to date with our tests. This bumps to the newest client-spec.


sighphyre (approved):
LGTM

PR #244, opened 2024-07-25
feat: Added new metadata fields to Registration and Metrics

chriswk:
https://linear.app/unleash/issue/2-2388/extend-java-sdk-with-new-metrics
https://linear.app/unleash/issue/2-2381/extend-c-sdk-with-new-metrics

### DIFF
--- a/src/main/java/io/getunleash/metric/ClientRegistration.java
+++ b/src/main/java/io/getunleash/metric/ClientRegistration.java
@@ -2,6 +2,7 @@ package io.getunleash.metric;
 
 import io.getunleash.event.UnleashEvent;
 import io.getunleash.event.UnleashSubscriber;
+import io.getunleash.lang.Nullable;
 import io.getunleash.util.UnleashConfig;
 import java.time.LocalDateTime;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


src/main/java/io/getunleash/metric/ClientRegistration.java
```diff
@@ -23,6 +28,10 @@ public class ClientRegistration implements UnleashEvent {
         this.started = started;
         this.strategies = strategies;
         this.interval = config.getSendMetricsInterval();
+        this.specVersion = config.getClientSpecificationVersion();
+        this.platformName = System.getProperty("java.vm.name", "JRE");
+        this.platformVersion = System.getProperty("java.vm.version", "1.8");
```
chriswk, line 33: What should our defaults be if we can't resolve these. Since we know we're requiring Java 1.8, I thought saying 1.8 was safe, but for the platform name (not really sure)

chriswk:
Running with experimental metrics and testing this change
```client_sdk_versions{sdk_name="unleash-client-java",sdk_version="9.2.3-SNAPSHOT",platform_name="OpenJDK 64-Bit Server VM",platform_version="17.0.11+9",yggdrasil_version="not-set",spec_version="5.1.5"} 1```

src/main/java/io/getunleash/metric/ClientRegistration.java
sighphyre: The Unleash side will handle nulls and transform them into something safe so I think null is a better choice

chriswk: Handled :)

sighphyre (approved):
LGTM

Issue #246, opened 2024-07-26
Wrong toggle evaluation with dependencies


DAGRSAG:
### Describe the bug

Hi,
we have the following config defined in our bootstrap file

```JSON
{
    "version": 2,
    "features": [
        {
            "name": "SR26_Pack",
            "enabled": false
        },
        {
            "name": "featureOne",
            "enabled": true,
            "dependencies": [
                {
                    "feature": "SR26_Pack"
                }
            ]
        }
    ]
}
```

My purpose here is, that we have toggle `featureOne` which should only be true, when the parent `SR26_Pack` is true.

If your code evaluates the toggle now, the return value is `true`, which is not expected.

In my opinion the problem is in this `DefaultUnleash` code here

```Java
UnleashContext enhancedContext = context.applyStaticFields(config);
if (featureToggle == null) {
    return new FeatureEvaluationResult(
            fallbackAction.test(toggleName, enhancedContext), defaultVariant);
} else if (!featureToggle.isEnabled()) {
    return new FeatureEvaluationResult(false, defaultVariant);
} else if (featureToggle.getStrategies().isEmpty()) {
    return new FeatureEvaluationResult(
            true, VariantUtil.selectVariant(featureToggle, context, defaultVariant));
}
```

this code means, that if no strategy is defined (`featureToggle.getStrategies().isEmpty()` returns `true`) always true is returned as evaluation value for the toggle.
The parent dependencies are only looked at for the first time in the else block afterwards.

Why not checking first if the `isParentDependencySatisfied` after `!featureToggle.isEnabled()`? Because if the parent dependencies not satisfied to toggle can not be true.

The behavior is correct, if e.g. the default strategy is added to the toggle def

### Expected behavior

Parent dependencies gets considered also if no strategy is defined

### Subscription type

Open source

### Hosting type

Self-hosted

### SDK information (language and version)

Java in version 9.2.0

DAGRSAG:
I should add, that we use the sdk in an offline mode. Means only the bootstrap file is loaded and decides about the toggle state

chriswk:
Hi @DAGRSAG . Thanks for your report. I think you're right; I'll see if I can't release a fix this week.

DAGRSAG:
Hi @chriswk,
thanks for the update and working on a fix.

DAGRSAG:
Awesome. Thanks. I will test it on Thursday 👍

PR #248, opened 2024-07-29
fix: Always check parent dependency if present

chriswk:
Fixes: #246

### DIFF
--- a/src/main/java/io/getunleash/DefaultUnleash.java
+++ b/src/main/java/io/getunleash/DefaultUnleash.java
@@ -179,56 +179,53 @@ public class DefaultUnleash implements Unleash {
                     fallbackAction.test(toggleName, enhancedContext), defaultVariant);
         } else if (!featureToggle.isEnabled()) {
             return new FeatureEvaluationResult(false, defaultVariant);
-        } else if (featureToggle.getStrategies().isEmpty()) {
-            return new FeatureEvaluationResult(
-                    true, VariantUtil.selectVariant(featureToggle, context, defaultVariant));
-        } else {
+        } else if (isParentDependencySatisfied(featureToggle, context, fallbackAction)) {
             // Dependent toggles, no point in evaluating child strategies if our dependencies are
             // not satisfied
-            if (isParentDependencySatisfied(featureToggle, context, fallbackAction)) {
-                for (ActivationStrategy strategy : featureToggle.getStrategies()) {
-                    Strategy configuredStrategy = getStrategy(strategy.getName());
-                    if (configuredStrategy == UNKNOWN_STRATEGY) {
-                        LOGGER.warn(
-                                "Unable to find matching strategy for toggle:{} strategy:{}",
-                                toggleName,
-                                strategy.getName());
-                    }
+            if (featureToggle.getStrategies().isEmpty()) {
+                return new FeatureEvaluationResult(
+                        true, VariantUtil.selectVariant(featureToggle, context, defaultVariant));
+            }
+            for (ActivationStrategy strategy : featureToggle.getStrategies()) {
+                Strategy configuredStrategy = getStrategy(strategy.getName());
+                if (configuredStrategy == UNKNOWN_STRATEGY) {
+                    LOGGER.warn(
+                            "Unable to find matching strategy for toggle:{} strategy:{}",
+                            toggleName,
+                            strategy.getName());
+                }
 
-                    FeatureEvaluationResult result =
-                            configuredStrategy.getResult(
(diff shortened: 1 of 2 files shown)


src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -179,56 +179,53 @@ private FeatureEvaluationResult getFeatureEvaluationResult(
...
-            if (isParentDependencySatisfied(featureToggle, context, fallbackAction)) {
-                for (ActivationStrategy strategy : featureToggle.getStrategies()) {
-                    Strategy configuredStrategy = getStrategy(strategy.getName());
-                    if (configuredStrategy == UNKNOWN_STRATEGY) {
-                        LOGGER.warn(
-                                "Unable to find matching strategy for toggle:{} strategy:{}",
-                                toggleName,
-                                strategy.getName());
-                    }
+            if (featureToggle.getStrategies().isEmpty()) {
```
gastonfournier, line 185: Should we do something like: 
```java
processedStrategies = featureToggle.getStrategies().map(this::getStrategy).filter(s -> s != UNKNOWN_STRATEGY);
if (processedStrategies.isEmpty()) {
...
```
?

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -179,56 +179,53 @@ private FeatureEvaluationResult getFeatureEvaluationResult(
...
-                                            featureToggle, context, defaultVariant);
-                        }
-                        result.setVariant(variant);
-                        return result;
+                FeatureEvaluationResult result =
+                        configuredStrategy.getResult(
+                                strategy.getParameters(),
+                                enhancedContext,
+                                ConstraintMerger.mergeConstraints(featureRepository, strategy),
+                                strategy.getVariants());
```
gastonfournier, line 203: I find it awkward that if we don't find strategies we return `enabled=true` but if we do find strategies but we can't map the strategy to a known strategy we (probably) return `enabled=false` at line 216

chriswk: I don't disagree, it's not because of this PR though.

Maybe a rethink here; as a separate thing. I'm open to some sparring

src/main/java/io/getunleash/DefaultUnleash.java
chriswk: Rethinking this is probably a good idea, but it doesn't relate to this PR, so I'd suggest leaving it for a separate sparring + PR session

src/main/java/io/getunleash/DefaultUnleash.java
sighphyre: > I find it awkward that if we don't find strategies we return enabled=true but if we do find strategies but we can't map the strategy to a known strategy we (probably) return enabled=false at line 216

I think you have to take this up with v0.0.1 of the SDK spec. That's the correct behavior. An empty list of strategies is valid. A custom strategy that you defined server side but didn't implement client side is a broken and is expected to short circuit to false

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -179,56 +179,53 @@ private FeatureEvaluationResult getFeatureEvaluationResult(
...
+                    if (variant == null) {
+                        variant = VariantUtil.selectVariant(featureToggle, context, defaultVariant);
                     }
+                    result.setVariant(variant);
+                    return result;
                 }
             }
-            return new FeatureEvaluationResult(false, defaultVariant);
         }
+        return new FeatureEvaluationResult(false, defaultVariant);
```
sighphyre, line 216: Something feels very subtly off here but it's not breaking tests sooo... it might be okay?

What's going through my head? Previously, we'd fall through to true, now we fall through to false. In theory, a toggle that has no strategies should fallback to the 'enabled' property, not true or false. That's distinct from a toggle that doesn't have strategies defined, which should never happen server side but when it does we turn off the toggle

So really we're into undefined territory here. It's probably okay but there's nothing in the client spec that enforces this behavior, so it's possible we break this again subtly

src/main/java/io/getunleash/DefaultUnleash.java
DAGRSAG: > A custom strategy that you defined server side but didn't implement client side is a broken and is expected to short circuit to false

I am completely with @sighphyre here. If a strategy is not found, the safe and correct way is to return false for the toggle if the strategy is not found

gastonfournier (approved):
Agreed this is a good step forward and we can iterate on the comments :+1:

PR #249, opened 2024-07-29
Setup coveralls action instead of ancient coveralls maven plugin

chriswk:
Bumps dependencies to fix some security warnings. Sets up to use coverallsaction instead of outdated coveralls maven plugin


gastonfournier (approved):
LG!

PR #255, opened 2024-11-19
UnleashContext.Builder Convenience

mxsb:
- made builder populate special properties by name

## About the changes
Making the io.getunleash.UnleashContext.Builder populate special properties for convenience. 
Sometimes users may specify a property by name and not by method. Currently this property will not be picked up.
This PR addresses this and picks up the special properties for the convenience of the user.


chriswk (approved):
Thank you for this.

It's a nice improvement. Will merge and release a patch version

Issue #234, opened 2023-12-07
Remove (again) our deprecated hashing methods


chriswk:
### Describe the feature request

To avoid users complaining about not seeing full/even distribution we should remove the methods that were added in 9.2.0 for simplifying the upgrade path from 8.x to 9.x. Ideally in 10.x but at the latest at 11.x

### Background

In #232 an active user complained that the upgrade from 8.3.1 to 9.x broke the way they worked, so in 9.2.0 we added methods allowing to use the old seed for the hashing of the variants.

### Solution suggestions

As the title says, we should remove the deprecated methods as soon as possible, to avoid having more users falling down the rabbit hole of having a 10% allocation getting the feature and then not seeing a 50/50 split between two variants of 50% / 50% weight.

hcgoranson:
Follow up from https://github.com/Unleash/unleash-client-java/issues/232: the last team on our side that was affected by this issue have now stopped using the deprecated seeding approach, so it might be time to remove to go ahead and remove it again

sighphyre:
@hcgoranson That is absolutely incredible timing, I was here to double check whether or not it was possible to remove these methods by pure chance.

Thanks for the update! They'll be leaving in the next major version!

PR #256, opened 2024-11-28
Yggdrasil engine

sighphyre:
Swaps out the evaluation of feature toggles for Yggdrasil

## Notable changes

- Previously the SDK would emit events for parent toggles when the child was evaluated. Looks like this SDK was doing something unusual. There's no longer a reasonable path to doing that since the SDK is no longer aware of the parent-child relationship, so it no longer does that
- Metrics is now done purely through Yggdrasil. The abstraction on the metrics service + Unleash interface for counting toggles and variants has been removed. Yggdrasil now acts both as sink and source of metrics
- Deprecated logic legacy variant distribution is removed. The plan was to remove this in the next major. That's now. User has confirmed on the open issue that they no longer need that behavior 
- Custom strategies are now handled by Yggdrasil via an adapter 

## To do

- The Yggdrasil engine doesn't yet support feature_enabled, so the evaluation for variants needs to resolve the toggle so that it can calculate metrics
- Platform classifier for working out which Yggdrasil jar to pull. Currently hardcoded to Linux x86_64 in the POM 
- MoreOperations needs to get trimmed. Direct access to feature toggle objects needs to get dropped
- A lot of removal of domain objects


pom.xml
```diff
@@ -56,6 +56,12 @@
   </scm>
 
     <dependencies>
+        <dependency>
+            <groupId>io.getunleash</groupId>
+            <artifactId>yggdrasil-engine</artifactId>
+            <version>0.1.0-alpha.9</version>
+            <classifier>x86_64-linux</classifier>
```
sighphyre, line 63: Classifier needs to be fixed before releasing this

src/main/java/io/getunleash/FakeUnleash.java
```diff
@@ -107,17 +107,6 @@ public void disableAllExcept(String... excludedFeatures) {
         }
     }
 
-    @Override
-    public Variant deprecatedGetVariant(String toggleName, UnleashContext context) {
```
sighphyre, line 111: This is a public interface change. This was in place to support legacy evaluation of variants. We're removing that in the next major, as per https://github.com/Unleash/unleash-client-java/issues/234

src/main/java/io/getunleash/FakeUnleash.java
```diff
@@ -176,15 +165,5 @@ public List<EvaluatedToggle> evaluateAllToggles(@Nullable UnleashContext context
                                             getVariant(toggleName)))
                     .collect(Collectors.toList());
         }
-
-        @Override
-        public void count(String toggleName, boolean enabled) {
```
sighphyre, line 181: Public interface change. I don't see a reason why anyone would be using this but it is public. The Yggdrasil engine is now the sink for this data

src/main/java/io/getunleash/metric/DefaultHttpMetricsSender.java
```diff
@@ -33,6 +35,7 @@ public DefaultHttpMetricsSender(UnleashConfig unleashConfig) {
         this.gson =
                 new GsonBuilder()
                         .registerTypeAdapter(LocalDateTime.class, new DateTimeSerializer())
+                        .registerTypeAdapter(Instant.class, new InstantSerializer())
```
sighphyre, line 38: Yggdrasil deals in Instants not Datetimes. This allows us to serialize those to ISO 8601 when we send metrics

src/main/java/io/getunleash/metric/ClientMetrics.java
```diff
@@ -1,5 +1,6 @@
 package io.getunleash.metric;
 
+import io.getunleash.engine.MetricsBucket;
```
gastonfournier, line 3: Probably not needed

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
```diff
@@ -50,29 +58,24 @@ public void register(Set<String> strategies) {
...
+                    int statusCode = metricSender.sendMetrics(metrics);
+                    if (statusCode >= 200 && statusCode < 400) {
+                        throttler.decrementFailureCountAndResetSkips();
+                    }
+                    if (statusCode >= 400) {
+                        throttler.handleHttpErrorCodes(statusCode);
+                    }
+                } catch (YggdrasilError e) {
+                    LOGGER.error(
+                            "Failed to retrieve metrics from the engine, this is a serious error",
```
gastonfournier, line 77:
```suggestion
                            "Failed to retrieve metrics from the engine, this is a serious error, please report it",
```

gastonfournier (approved):
Looking great!

src/main/java/io/getunleash/metric/ClientMetrics.java
sighphyre: Its a bit sneaky but this was a drop in replacement for the existing metrics bucket. All I needed to do was include the new import and the code is unchanged

pom.xml
sighphyre: This is now fixed

PR #259, opened 2025-01-08
feat/yggdrasil feature enabled

sighphyre:
Forces the Java SDK to resolve feature enabled from Yggdrasil rather than computing it separately from isEnabled + getVariant


src/test/java/io/getunleash/integration/TestCaseVariant.java
```diff
@@ -22,7 +22,14 @@ public String getToggleName() {
 
     public Variant getExpectedResult() {
         if (expectedResult.getName().equals("disabled")) {
-            return Variant.DISABLED_VARIANT;
+            Variant clone =
+                    new Variant(
```
sighphyre, line 26: This seems really dumb, is there a better way to clone an object?

pom.xml
```diff
@@ -10,7 +10,7 @@
         <version.junit5>5.10.3</version.junit5>
         <version.okhttp>4.12.0</version.okhttp>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <version.unleash.specification>5.1.5</version.unleash.specification>
+        <version.unleash.specification>5.1.9</version.unleash.specification>
```
gastonfournier, line 13: Does it work well with https://github.com/Unleash/unleash-client-java/pull/260 where you had to roll down?

src/test/java/io/getunleash/integration/TestCaseVariant.java
gastonfournier: You can have a clone method that does this for you, but the code will be there... I remember there were some utility methods, but atm I think this is fine

gastonfournier (commented):
LGTM!

pom.xml
sighphyre: It does not! It's on the list of things to do prior to release - a proper set of tests around Windows + unicode. But this is okay for now so these PRs don't get too big I think

PR #260, opened 2025-01-09
chore/os-classifiers-for-build

sighphyre:
Adds Maven classifiers to the build system so that maven automatically picks up the correct os/arch at build time and installs the correct version of Yggdrasil that contains that platform's binary

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>io.getunleash</groupId>
             <artifactId>yggdrasil-engine</artifactId>
-            <version>0.1.0-alpha.10</version>
+            <version>0.1.0-alpha.11</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -10,7 +10,7 @@
         <version.junit5>5.10.3</version.junit5>
         <version.okhttp>4.12.0</version.okhttp>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <version.unleash.specification>5.1.9</version.unleash.specification>
+        <version.unleash.specification>5.1.7</version.unleash.specification>
```
sighphyre, line 13: Rolled this back because Windows needs a patch to handle unicode

gastonfournier (approved):
LGTM! Let's test

PR #261, opened 2025-01-15
chore: update yggdrasil to handle latest spec

sighphyre:
Bumps Yggdrasil to expose the core version and handle UTF-8 characters

"Fixes" a timing issue on Windows in the tests. How did we not notice that before? Well we've never run these tests on a Windows runner

### DIFF
--- a/src/test/java/io/getunleash/DefaultUnleashTest.java
+++ b/src/test/java/io/getunleash/DefaultUnleashTest.java
@@ -310,6 +310,7 @@ class DefaultUnleashTest {
                         .build();
 
         Unleash unleash = new DefaultUnleash(config);
+        Thread.sleep(1);
         verify(fetcher, times(1)).fetchFeatures();
         Thread.sleep(1200);
         verify(fetcher, times(2)).fetchFeatures();
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
         <version.junit5>5.10.3</version.junit5>
         <version.okhttp>4.12.0</version.okhttp>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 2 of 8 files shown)


src/test/java/io/getunleash/DefaultUnleashTest.java
```diff
@@ -310,6 +310,7 @@ public void asynchronous_fetch_on_initialisation_fails_silently_and_retries()
                         .build();
 
         Unleash unleash = new DefaultUnleash(config);
+        Thread.sleep(1);
```
sighphyre, line 313: *cries in Windows* So this is the failing test that sent me down the rabbit hole of the client spec tests (which aren't yet fixed). It looks like there's a very subtle difference in the thread scheduler between Windows and Linux. Linux is more precise, which means that the fetch has always completed before the verify runs on the next line. On Windows, this never completes without this sleep

gastonfournier: Yes, we can't wait on DefaultUnleash constructor, but tbh, I don't like it does `fetchFeatures` as part of constructing the object... I think fetch features should be part of some `initialize()` method and then maybe you'd be able to wait on it... but sleeping is fine, people need to sleep :)

gastonfournier (approved):
I like it!

src/test/java/io/getunleash/DefaultUnleashTest.java
sighphyre: Agreed, I'd like it to change too and what you propose is what I'd like to do too, Constructors should construct, not spawn threads.

Small steps though, this is legacy behaviour that's being surfaced because we're doing Windows tests now. Can fix next series of PRs

PR #263, opened 2025-01-20
feat: client identification headers

kwasniew:
## About the changes

This PR adds client identification headers to the feature and metrics calls that the client makes to Unleash. The headers are:
- `x-unleash-appname`: the name of the application that is using the client. `UNLEASH-APPNAME` will be deleted in another PR (expand/contract pattern)
- `x-unleash-connection-id`: an internal unique identifier for the current instance of the client generated by the built-in UUID lib
- `x-unleash-sdk`: sdk information in the format `unleash-java@<version>` following reference `unleash-lang@<version>` naming pattern

Reference implementation: https://github.com/Unleash/unleash-client-node/pull/690

## Discussion points

There's a failing test that I can't reproduce failing locally.

### DIFF
--- a/src/main/java/io/getunleash/util/UnleashConfig.java
+++ b/src/main/java/io/getunleash/util/UnleashConfig.java
@@ -221,7 +221,7 @@ public class UnleashConfig {
         return instanceId;
     }
 
-    public String getConnectionId() {
+    String getConnectionId() {
         return connectionId;
     }
 
--- a/src/main/java/io/getunleash/util/UnleashConfig.java
+++ b/src/main/java/io/getunleash/util/UnleashConfig.java
@@ -23,12 +23,16 @@ import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.function.Consumer;
 
 public class UnleashConfig {
 
-    public static final String UNLEASH_APP_NAME_HEADER = "UNLEASH-APPNAME";
+    public static final String LEGACY_UNLEASH_APP_NAME_HEADER = "UNLEASH-APPNAME";
     public static final String UNLEASH_INSTANCE_ID_HEADER = "UNLEASH-INSTANCEID";
+    public static final String UNLEASH_CONNECTION_ID_HEADER = "X-UNLEASH-CONNECTION-ID";
+    public static final String UNLEASH_APP_NAME_HEADER = "X-UNLEASH-APPNAME";
+    public static final String UNLEASH_SDK_HEADER = "X-UNLEASH-SDK";
 
     private final URI unleashAPI;
     private final UnleashURLs unleashURLs;
@@ -37,6 +41,7 @@ public class UnleashConfig {
     private final String appName;
     private final String environment;
     private final String instanceId;
+    private final String connectionId;
     private final String sdkVersion;
     private final String backupFile;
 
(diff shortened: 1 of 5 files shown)


src/test/java/io/getunleash/util/UnleashConfigTest.java
```diff
@@ -114,7 +113,7 @@ public void should_set_sdk_version() {
         UnleashConfig config =
                 UnleashConfig.builder().appName("my-app").unleashAPI("http://unleash.org").build();
 
-        assertThat(config.getSdkVersion()).isEqualTo("unleash-client-java:development");
```
kwasniew, line 117: I don't think we're using it anywhere so should be safe to rename it to a common convention with other SDKs. The convention is inspired by npm versions since Node is a dominant part of our stack.

src/main/java/io/getunleash/util/UnleashConfig.java
```diff
@@ -716,7 +736,7 @@ public String getDefaultSdkVersion() {
             String version =
                     Optional.ofNullable(getClass().getPackage().getImplementationVersion())
                             .orElse("development");
-            return "unleash-client-java:" + version;
```
kwasniew, line 719: I don't think we're using it anywhere so should be safe to rename it to a common convention with other SDKs. The convention is inspired by npm versions since Node is a dominant part of our stack.

src/main/java/io/getunleash/util/UnleashConfig.java
```diff
@@ -449,6 +464,10 @@ static String getDefaultInstanceId() {
             return getHostname() + "generated-" + Math.round(Math.random() * 1000000.0D);
         }
 
+        static String getDefaultConnectionId() {
+            return UUID.randomUUID().toString();
```
kwasniew, line 468: this is not UUID v4 but it's the most idiomatic UUID I found in vanilla Java and we also use it for event ids.

src/main/java/io/getunleash/util/UnleashConfig.java
sighphyre: Yeah I also think that's fine

src/main/java/io/getunleash/util/UnleashConfig.java
```diff
@@ -211,6 +221,10 @@ public String getInstanceId() {
         return instanceId;
     }
 
+    public String getConnectionId() {
```
sighphyre, line 224: I'm probably committing some terrible Java sin here but I'd be tempted to remove the public access modifier here. We've been burned so many times by exposing things that we didn't need to and having someone form a weird dependency on them
```suggestion
    String getConnectionId() {
```

sighphyre (approved):
Nice, LGTM

sighphyre:
So I see the tests are failing and I see it's our old friend `asynchronous_fetch_on_initialisation_fails_silently_and_retries`. This test was failing for me on the Yggdrasil swap out branch (I honestly assumed it was something I broke in that branch), but I see it's having a similar problem here. I had to change the end lines in the test:

```
        Unleash unleash = new DefaultUnleash(config);
        verify(fetcher, times(1)).fetchFeatures();
        Thread.sleep(1200);
        verify(fetcher, times(2)).fetchFeatures();
```

To this:

```
        Unleash unleash = new DefaultUnleash(config);
        Thread.sleep(1); // new sleep
        verify(fetcher, times(1)).fetchFeatures();
        Thread.sleep(1200);
        verify(fetcher, times(2)).fetchFeatures();
```
That's probably fine here if you can't get that test to pass

src/main/java/io/getunleash/util/UnleashConfig.java
kwasniew: Ah good catch. Forgot about package-private access in Java :)

PR #264, opened 2025-01-21
fix: revert previous version scheme

kwasniew:
## About the changes

After some digging I realized that same version setting we use for the new `x-unleash-sdk` is also used for the registration calls. I'm reverting back the `unleash-client-java:version` convention instead of the new `unleash-java@version`. We use the colon convention a lot in the backend code. Since the new header is not used yet it can be synced with the old convention for now.


sighphyre (approved):
Ah good spot

PR #265, opened 2025-01-22
chore: Rework internal Yggdrasil abstractions

sighphyre:
Sorry this is so big, it turns out fixing the race condition in loading the bootstrap, required a new abstraction and that either required updating all the legacy class files or throwing them out. I chose to throw them out

# List of changes

## FeatureRepository 

The concrete implementation of this class is now the only place where `takeState` can be called on the Unleash engine, giving a single hook for changes. This means bootstrap, backup and update features are all handled through this class and the results of that are passed to the engine directly rather than using reflection to set the engine state externally. The public methods on this class are now purely to expose access to the UnleashEngine's public API. 

## Unleash Metric Service

This now exposes the metrics methods on the UnleashEngine `countToggle` and `countVariant`. Similarly to FeatureRepository, it acts as a wrapper between the existing Java code and the UnleashEngine's public API.

## Engine Proxy

This is a new abstraction that acts as a decorating proxy which bridged both the MetricsServiceImpl and the FeatureRepositoryImpl. This is done purely because both the metrics service and the feature repository need to close over the new UnleashEngine class. And they really, really should be the same UnleashEngine. Having this class in place means DefaultUnleash can't be created with the wrong engine in the wrong places.

EngineProxy is an interface implementing both MetricsService and FeatureRepository interfaces and the concrete EngineProxyImpl requires both a MetricsServiceImpl and a FeatureRepositoryImpl. This means the EngineProxy can be used transparently in place of either the FeatureRepository or the MetricsService and simply forwards calls onto the relevant concrete inplementation.

## Bootstrap/Backup

All methods that return a client feature set, for example, the bootstrapping API. Now return `Optional<String>`. Either they contain a valid client feature set or they return a None Option. All legacy data structures related to client features are now purely Yggdrasil's business.

## Events

Events that historically returned a set of parsed client features can obviously no longer do that since those objects no longer exist/ These events now contain the pure String set of those. They do expose a method that will lazily parse+cache a set of FeatureDefs - a lightweight object containing the name, type and project of the toggle rather than the whole definition

The Event constructor for ClientFeaturesResponse is now private. An appropriate event can be constructed using the static methods updated/unavailable/notChanged which will not allow you to construct the event without the correct data. This was done because these events were constructed incorrectly in the codebase in a lot of places and it was breaking tests.

## More Operations

More operations now likewise returns a List of FeatureDefs rather than the whole client feature set. Again because that object no longer exists

## Deprecated Removals

Anything marked as deprecated has been removed


src/main/java/io/getunleash/FeatureDefinition.java
```diff
@@ -0,0 +1,35 @@
...
+import io.getunleash.engine.FeatureDef;
+import java.util.Optional;
+
+public class FeatureDefinition {
+
+    private final String name;
+    private final Optional<String> type;
+    private final String project;
+
+    public FeatureDefinition(FeatureDef source) {
```
gastonfournier, line 12: Isn't FeatureDef and FeatureDefinition exactly the same? If so we could drop one of the two

src/main/java/io/getunleash/event/ClientFeaturesResponse.java
```diff
@@ -0,0 +1,83 @@
...
+        return httpStatusCode;
+    }
+
+    public Status getStatus() {
+        return statusCode;
+    }
+
+    public List<FeatureDefinition> getFeatures() {
+        if (clientFeatures.isPresent() && features == null) {
+            features = ClientFeaturesParser.parse(clientFeatures.get());
```
gastonfournier, line 61: Why not parsing in the constructor and making getFeatures just a getter?

src/main/java/io/getunleash/event/FeatureSet.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public FeatureSet(String clientFeatures) {
+        this.clientFeatures = clientFeatures;
+    }
+
+    public List<FeatureDefinition> getFeatures() {
+        if (features == null) {
+            features = ClientFeaturesParser.parse(clientFeatures);
+        }
+        return features;
```
gastonfournier, line 20: Same as above, I'd like to see the getter just doing get

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -0,0 +1,219 @@
...
+
+        if (unleashConfig.isSynchronousFetchOnInitialisation()) {
+            if (this.unleashConfig.getStartupExceptionHandler() != null) {
+                updateFeatures(this.unleashConfig.getStartupExceptionHandler()).run();
+            } else {
+                updateFeatures(
+                                e -> {
+                                    throw e;
+                                })
+                        .run();
```
gastonfournier, line 117:
```suggestion
                updateFeatures(
                                e -> {
                                    throw e;
                                }) // just throw exception handler
                        .run();
```

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -0,0 +1,219 @@
...
+    private void initCollections(UnleashScheduledExecutor executor) {
+        Optional<String> features = this.featureBackupHandler.read();
+        if (!features.isPresent() && this.bootstrapper != null) {
+            features = this.bootstrapper.read();
+        }
+        if (features.isPresent()) {
+            try {
+                this.engine.takeState(features.get());
+            } catch (YggdrasilInvalidInputException | YggdrasilError e) {
+                LOGGER.error("Error when initializing feature toggles", e);
```
gastonfournier, line 105: Should this emit an event?

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    @Override
+    public Optional<VariantDef> getVariant(String toggleName, UnleashContext context) {
+        try {
+            return Optional.ofNullable(
+                    this.engine.getVariant(toggleName, YggdrasilAdapters.adapt(context)));
+        } catch (YggdrasilInvalidInputException | YggdrasilError e) {
+            LOGGER.error("Error when checking feature toggle {}", toggleName, e);
+            return null;
```
gastonfournier, line 196: Similar comment as above
```suggestion
            return Optional.empty();
```
Or maybe default variant

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -0,0 +1,219 @@
...
+        return this.throttler.getSkips();
+    }
+
+    @Override
+    public Boolean isEnabled(String toggleName, UnleashContext context) {
+        try {
+            return this.engine.isEnabled(toggleName, YggdrasilAdapters.adapt(context));
+        } catch (YggdrasilInvalidInputException | YggdrasilError e) {
+            LOGGER.error("Error when checking feature toggle {}", toggleName, e);
+            return null;
```
gastonfournier, line 185: Do we want to return null or just false? I know this would be a major issue... but between returning null and just throwing I believe I err towards throwing, cause a NullPointerException would be confusing, but I believe the option I like the most is:

```suggestion
            return false; // if we had the default value for the feature here, that'd be even better
```

src/main/java/io/getunleash/repository/HttpFeatureFetcher.java
```diff
@@ -59,30 +61,24 @@ private ClientFeaturesResponse getFeatureResponse(
...
-                    ClientFeaturesResponse.Status.UNAVAILABLE,
-                    responseCode,
-                    getLocationHeader(request));
+            return ClientFeaturesResponse.unavailable(responseCode, getLocationHeader(request));
         }
     }
 
     private ClientFeaturesResponse followRedirect(HttpURLConnection request) throws IOException {
-        String newUrl = getLocationHeader(request);
+        String newUrl = getLocationHeader(request).get();
```
gastonfournier, line 81: If null this will fail hard, just details:
```suggestion
        String newUrl = getLocationHeader(request).orElseThrow(() -> new IllegalStateException("No Location header found in redirect response."));
```

gastonfournier (approved):
It looks good to me, I've left a few comments which I consider minor details, we can talk about them later but overall looks like an improvement, especially removing nulls and using Optional

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -229,38 +139,23 @@ public Variant getVariant(String toggleName, Variant defaultValue) {
...
+        UnleashContext enhancedContext = context.applyStaticFields(config);
 
-        if (strategies != null) {
-            for (Strategy strategy : strategies) {
-                map.put(strategy.getName(), strategy);
-            }
+        Variant variant =
+                YggdrasilAdapters.adapt(
+                        this.featureRepository.getVariant(toggleName, enhancedContext),
+                        defaultValue);
```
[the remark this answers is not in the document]
sighphyre: While I think your suggestion is probably more correct semantically, this makes testing the DefaultUnleash layer incredibly difficult because there's no way for the mock to signal to the Unleash layer that no toggle was found for that value.

I can try reworking it but I kinda value the testability that this adds (there's actually a really badly named commit that moved this responsibility). I'd like to leave it if you're okay with it

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -292,13 +193,19 @@ public List<EvaluatedToggle> evaluateAllToggles() {
...
-                                boolean enabled =
-                                        isEnabled(toggleName, context, (name, ctx) -> false);
-                                Variant variant = getVariant(toggleName, context, DISABLED_VARIANT);
-                                return new EvaluatedToggle(toggleName, enabled, variant);
+                                UnleashContext enhancedContext = context.applyStaticFields(config);
+                                Variant variant =
+                                        YggdrasilAdapters.adapt(
+                                                featureRepository.getVariant(
+                                                        toggleName, enhancedContext),
+                                                DISABLED_VARIANT);
```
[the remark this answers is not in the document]
sighphyre: Addressed in previous comment (I think)

src/main/java/io/getunleash/FeatureDefinition.java
```diff
@@ -0,0 +1,35 @@
+package io.getunleash;
+
+import io.getunleash.engine.FeatureDef;
+import java.util.Optional;
+
+public class FeatureDefinition {
+
+    private final String name;
+    private final Optional<String> type;
```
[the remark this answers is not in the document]
sighphyre: Your instincts are perfect, as usual. This is legacy on two layers. One in the Dotnet SDK itself (originating from [here](https://github.com/Unleash/unleash-client-dotnet/issues/66)), we specifically added a way for the user to check the type of the toggle and this is me trying to keep the SDKs consistent.

The second legacy layer is on Unleash itself, we stopped sending 'type' a year ago but there will still be Unleash instances in the wild that do

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -229,38 +139,23 @@ public Variant getVariant(String toggleName, Variant defaultValue) {
...
-            for (Strategy strategy : strategies) {
-                map.put(strategy.getName(), strategy);
-            }
+        Variant variant =
+                YggdrasilAdapters.adapt(
+                        this.featureRepository.getVariant(toggleName, enhancedContext),
+                        defaultValue);
+        this.metricService.countToggle(toggleName, variant.isFeatureEnabled());
+        this.metricService.countVariant(toggleName, variant.getName());
+        eventDispatcher.dispatch(new ToggleEvaluated(toggleName, variant.isEnabled()));
```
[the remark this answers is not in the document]
sighphyre: No this is a bug. Suggestion taken. Fantastic spot, dude

src/main/java/io/getunleash/FeatureDefinition.java
sighphyre: They contain the same data but FeatureDefinition is an Yggdrasil entity and I don't like leaking types from transient dependencies

src/main/java/io/getunleash/event/ClientFeaturesResponse.java
sighphyre: You may not need this data, for example if you're just checking that the SDK is actually polling. If you don't need the data itself the cost of a parse is expensive. The lazy initialization means you only pay that cost if you do need it and then only once

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
sighphyre: The empty response here is intentional. SDKs should handle missing toggles vs false toggles differently and that's in how the fallback function is executed (and for other SDKs how metrics are calculated). This lets us signal to the layer above that the toggle isn't here without having to leak the fallback handling into this layer

Could swap this for an Optional<boolean> if this bothers you though?

src/main/java/io/getunleash/DefaultUnleash.java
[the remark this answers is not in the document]
gastonfournier: I'm ok with it. I've found that this interface comes from the strategy and I think it's fine there because the strategy should just care about executing itself, the fallback value should be outside the strategy and maybe this is a consequence of that.

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
sighphyre: Added in https://github.com/Unleash/unleash-client-java/pull/265/commits/1331f3fcd009772a136aac0bbfa8321206b18b9a

src/main/java/io/getunleash/FeatureDefinition.java
gastonfournier: I was going to ask... is it? But I think you swap them as FeatureDef is an Yggdrasil entity. So yes, given that I think it's ok to have the two

src/main/java/io/getunleash/event/ClientFeaturesResponse.java
gastonfournier: Fair point!

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
gastonfournier: I think  it's fine as long as FeatureRepository is an internal interface (or meant to be internal). If it's exposed as something a client can use then I think optional is better, but in this case I understand the point and I think it's fine to return null

PR #268, opened 2025-01-26
#267 GraalVM Reachability Metadata

### LINKED ISSUE #267: graalvm reachability metadata

Ensure the Unleash Java client can be compiled to a native binary without manual configuration.

### Background

_No response_

### Solution suggestions

Either include the reachability data in the unleash-client-java or add a reachability modification into https://github.com/oracle/graalvm-reachability-metadata

iromu:
## About the changes

Closes #267


sighphyre:
I know very little about actually using Graal with Java but is it possible to include an example or a test for this? The reason I ask is because in the near future this library is going to take on a native dependency and I suspect that's going to break this.

This does look fine to me though

chriswk (approved):
Thanks for your contribution @iromu .

PR #269, opened 2025-01-29
docs: migration guide

sighphyre:
Adds some notes around how to migrate to the upcoming v10 version and what the user should look out for during the migration process


melindafekete (approved):
Nice, I think it's well-written and really clear.

PR #270, opened 2025-01-29
chore: add jmh benchmarks for sdk

chriswk:
Adds a couple of JMH benchmarks for the SDK. Stupidly simple for now, just to build a baseline before upgrading/migrating to Yggdrasil engine.


sighphyre (approved):
LGTM

PR #274, opened 2025-02-03
task: This sets up JReleaser, so we can run future releases from Github Actions instead of only from my machine

chriswk:


sighphyre (approved):
Yaaaay!

PR #289, opened 2025-03-03
fix: different interval headers are now sent for metrics/features

sjaanus:
Previously same interval was sent to both requests, metrics and features. Now for metrics we are sending metrics interval and for feature fetching, we send refreshInterval.

### DIFF
--- a/src/main/java/io/getunleash/repository/OkHttpFeatureFetcher.java
+++ b/src/main/java/io/getunleash/repository/OkHttpFeatureFetcher.java
@@ -21,10 +21,10 @@ import static io.getunleash.util.UnleashConfig.UNLEASH_INTERVAL;
 public class OkHttpFeatureFetcher implements FeatureFetcher {
     private final HttpUrl toggleUrl;
     private final OkHttpClient client;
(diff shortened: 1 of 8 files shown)


src/main/java/io/getunleash/repository/OkHttpFeatureFetcher.java
```diff
@@ -16,11 +16,15 @@
 import okhttp3.Request;
 import okhttp3.Response;
 
+import static io.getunleash.util.UnleashConfig.UNLEASH_INTERVAL;
+
 public class OkHttpFeatureFetcher implements FeatureFetcher {
     private final HttpUrl toggleUrl;
     private final OkHttpClient client;
+    private final UnleashConfig config;
```
kwasniew, line 24: do you need to store the entire config if you only need the interval?

kwasniew (approved):

src/main/java/io/getunleash/repository/OkHttpFeatureFetcher.java
sjaanus: Storing interval only now

PR #295, opened 2025-03-12
feat: add environmentEnabled property on feature defs

sighphyre:
Restores the old `enabled` property on feature defs. I've chosen to rename this to `environmentEnabled`, which I think expresses better what this actually means


chriswk (approved):
LRTM :)

Issue #266, opened 2025-01-24
Allow contexts to be used in unit testing with FakeUnleash


markusjevringflowers:
### Describe the feature request

I would like to be able to pass an `UnleashContext` to `FakeUnleash.enable()`, so that it matches `Unleash.isEnabled()`

### Background

You have the FakeUnleash for unit testing, which is great. However, for things like dependency injection, and per your documentation, you will often end up with *a singleton instance* of FakeUnleash. Especially when you are re-using the spring context between tests. If the tests are single-threaded, this fine, as you can mutate global state without fear of interaction. _However, with concurrent tests, mutating global state like this can lead to incorrect interactions_.

### Solution suggestions

Make `FakeUnleash` track and use `UnleashContext` so that `FakeUnleash.enable()` matches `Unleash.isEnabled()`

gastonfournier:
Thanks, it seems reasonable, if you'd like, feel free to open a PR with your suggestion.

markusjevringflowers:
I have a draft PR for this open. I will get some more eyes on it, and make sure it works correctly, and then I'll mark it ready.

markusjevringflowers:
Hello @gastonfournier, the PR is ready for review now: https://github.com/Unleash/unleash-client-java/pull/298

chriswk:
Approved and merged #298.

chriswk:
Released in 10.2.2, closing this issue. Thanks for both the issue and the great PR.

PR #298, opened 2025-03-21
Conditionally enable features based on contexts

markusjevringflowers:
## About the changes
Conditionally enable feature toggles based on matching the context. 
See https://github.com/Unleash/unleash-client-java/issues/266

### Important files
* FakeUnleash
* FakeUnleashTest

## Discussion points
We assume that we will want to define *multiple* conditions concurrently, because that's how unleash works in reality. As a result, `FakeUnleash.conditionallyEnable()` can be called multiple times, and these conditions are ORed together. That way one test can say "enable for x=1" and another says "enable for y=2", and these can co-exist without affecting each other.

We assume that enabling or disabling a feature toggle without specifying a context wipes out any conditionally enabled configurations for the same feature toggles, and vice versa. See the difference between `features` and `conditionalFeatures`

We also assume that we need to be able to do this concurrently in terms of execution, hence the use of concurrent classes in FakeUnleash.

### DIFF
--- a/src/main/java/io/getunleash/FakeUnleash.java
+++ b/src/main/java/io/getunleash/FakeUnleash.java
@@ -115,12 +115,14 @@ public class FakeUnleash implements Unleash {
 
     public void enable(String... features) {
         for (String name : features) {
+            this.conditionalFeatures.remove(name);
             this.features.put(name, true);
         }
     }
 
     public void disable(String... features) {
         for (String name : features) {
+            this.conditionalFeatures.remove(name);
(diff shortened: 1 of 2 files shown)


chriswk (approved):
Hi @markusjevringflowers . Great contribution. I'll get it merged. We have a bug that I want to fix, and then I'll release a new minor version of the SDK. Thank you!

markusjevringflowers:
@chriswk Thanks. I'm happy to have contributed! =)

Issue #288, opened 2025-02-27
Upgrading Unleash causes Rust thread panic


lbenedetto:
### Describe the bug

When we upgraded Unleash java client from version 9.2.6 to 10.0.2, we started seeing errors in production

We reverted the upgrade and the errors went away.

### Logs, error output, etc.

```bash
thread '<unnamed>' panicked at /cargo/registry/src/index.crates.io-6f17d22bba15001f/serde_json-1.0.138/src/ser.rs:2083:64:
byte index 2 is not a char boundary; it is inside '\t' (bytes 1..2) of `�	����nt`                                       
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace                                             
fatal runtime error: failed to initiate panic, error 4117039104             
```
Some of these errors contained recognizable flag names
```bash
byte index 4 is not a char boundary; it is inside '\u{4}' (bytes 3..4) of `�g��nd_profile_feature_card_notification_LC_enabled`
byte index 4 is not a char boundary; it is inside '\0' (bytes 3..4) of `�����NU_Highlighted_Item_Design`
```

### Unleash version

10.0.2

### Subscription type

None

### Hosting type

None

sighphyre:
So sorry about that, are there by any chance unicode strings in your contexts?

We'll take a look, thank you for reporting this

lbenedetto:
I'll have someone do a thorough check to make sure there aren't. But at least in the strings mentioned in the errors, there are not. For example, the full string for one of the errors is `CE_backend_profile_feature_card_notification_LC_enabled`

sighphyre:
Hey @lbenedetto, I can't seem to reproduce this at all. Any other context you can provide here? What OS are you running on? Any chance you've tinkered with encoding? Is this a heavy multi-threaded workload?

hcgoranson:
Hello, I talked to the person that created the `CE_backend_profile_feature_card_notification_LC_enabled` flag and it was done in the "usual way", so no difference compared to how we usually create flags and no special encoding related setup

We are running our Java application in a docker base image from  Amazon Corretto 21 and the flag is not exposed more multi-threads compared to our other flags

gastonfournier:
Hi @hcgoranson could you share with us an export of the feature configuration? If you prefer not to share it here, you can find me or Simon on our community Slack and you can send either or both of us a DM. 

I'd like to reproduce it on my own, and I'll try out some things but having the same configuration as you may save us a lot of time.

[...]

PR #304, opened 2025-04-01
chore: upgrade spring-boot-sample to unleash 10

gastonfournier:
This is an attempt to verify #288 with SpringBoot: `./gradlew bootRun` and accessing `http://localhost:8080/all` that evaluates all toggles at once.

The result was unsuccessful as I couldn't reproduce the issue. 

I also tried locally with a custom header and without synchronousFetchOnInitialisation:
```java
        UnleashConfig config = UnleashConfig.builder().unleashAPI(url).appName(appName)
            //.apiKey(apiKey)
            .customHttpHeader("Authorization", apiKey)
            .instanceId("spring-boot-example")
            .environment("development")
            //.synchronousFetchOnInitialisation(true)
            .fetchTogglesInterval(15).build();
```

Still the PR is valid to bump the examples

### DIFF
--- a/examples/spring-boot-example/src/main/java/io/getunleash/unleash/example/ProxyController.java
+++ b/examples/spring-boot-example/src/main/java/io/getunleash/unleash/example/ProxyController.java
@@ -1,10 +1,10 @@
 package io.getunleash.unleash.example;
 
 import io.getunleash.Unleash;
-import jakarta.websocket.server.PathParam;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.bind.annotation.PathVariable;
 
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


examples/spring-boot-example/src/main/java/io/getunleash/unleash/example/ProxyController.java
```diff
@@ -33,7 +33,7 @@ public Map<String, Boolean> getAllToggles() {
     }
 
     @GetMapping("/toggle/{toggleName}")
-    public Boolean getToggle(@PathParam("toggleName") String name) {
+    public Boolean getToggle(@PathVariable("toggleName") String name) {
```
gastonfournier, line 36: Correct way of getting the path variable

gastonfournier:
We're not going to upgrade to v10 as we're looking at alternatives to have [Yggdrasil](https://github.com/Unleash/yggdrasil) inside Java SDK which will likely come with v11

PR #306, opened 2025-06-04
chore: update to wasm engine

sighphyre:
Moves the internals to use the new Yggdrasil WASM engine. The WASM engine collapses isEnabled/getVariant, impressionData and metrics into a one shot call, so updates here to conform to that new API


chriswk (approved):
Yeah, this looks good to me.

PR #308, opened 2025-06-13
chore: remove java 8 build

gastonfournier:
Release 11 dropped support for Java 8. Removing that from main build steps


sighphyre (approved):
Thank you!

PR #313, opened 2025-07-04
chore: bump logback core version

sighphyre:
Closes a [vulnerability](https://github.com/Unleash/unleash-client-java/security/dependabot/15)


chriswk (approved):
Now that we require 11. Good

PR #318, opened 2025-08-07
feat: streaming support

kwasniew:
## About the changes

Closes #

### DIFF
--- a/src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
+++ b/src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
@@ -10,6 +10,8 @@ import io.getunleash.engine.YggdrasilInvalidInputException;
 import io.getunleash.event.ClientFeaturesResponse;
 import io.getunleash.event.EventDispatcher;
 import io.getunleash.event.UnleashReady;
+import io.getunleash.streaming.NoOpStreamingFeatureFetcher;
+import io.getunleash.streaming.StreamingFeatureFetcher;
 import io.getunleash.util.Throttler;
 import io.getunleash.util.UnleashConfig;
 import io.getunleash.util.UnleashScheduledExecutor;
@@ -25,10 +27,20 @@ public class FeatureRepositoryImpl implements FeatureRepository {
     private final BackupHandler featureBackupHandler;
     private final ToggleBootstrapProvider bootstrapper;
     private final FeatureFetcher featureFetcher;
+    private final StreamingFeatureFetcher streamingFeatureFetcher;
     private final EventDispatcher eventDispatcher;
     private final UnleashEngine engine;
     private final Throttler throttler;
     private boolean ready;
+
+    private StreamingFeatureFetcher createStreamingFeatureFetcher(
+            UnleashConfig config, Consumer<String> streamingUpdateHandler) {
+        if (config.isStreamingMode()) {
+            return new StreamingFeatureFetcherImpl(config, streamingUpdateHandler);
+        } else {
+            return new NoOpStreamingFeatureFetcher();
+        }
+    }
 
     public FeatureRepositoryImpl(UnleashConfig unleashConfig, UnleashEngine engine) {
         this(unleashConfig, new FeatureBackupHandlerFile(unleashConfig), engine);
@@ -36,23 +48,47 @@ public class FeatureRepositoryImpl implements FeatureRepository {
 
     public FeatureRepositoryImpl(
             UnleashConfig unleashConfig, BackupHandler featureBackupHandler, UnleashEngine engine) {
-        this(
-                unleashConfig,
-                featureBackupHandler,
-                engine,
-                unleashConfig.getUnleashFeatureFetcherFactory().apply(unleashConfig));
+        this(unleashConfig, featureBackupHandler, engine, new EventDispatcher(unleashConfig));
+    }
+
+    private FeatureRepositoryImpl(
+            UnleashConfig unleashConfig,
+            BackupHandler featureBackupHandler,
+            UnleashEngine engine,
+            EventDispatcher eventDispatcher) {
+        this.unleashConfig = unleashConfig;
+        this.featureBackupHandler = featureBackupHandler;
+        this.engine = engine;
+        this.featureFetcher = unleashConfig.getUnleashFeatureFetcherFactory().apply(unleashConfig);
+        this.bootstrapper = unleashConfig.getToggleBootstrapProvider();
+        this.eventDispatcher = eventDispatcher;
+        this.throttler =
+                new Throttler(
+                        (int) unleashConfig.getFetchTogglesInterval(),
+                        300,
+                        unleashConfig.getUnleashURLs().getFetchTogglesURL());
+
+        this.streamingFeatureFetcher = initializeStreamingFeatureFetcher();
+
+        this.initCollections(unleashConfig.getScheduledExecutor());
+    }
+
+    private StreamingFeatureFetcher initializeStreamingFeatureFetcher() {
+        return createStreamingFeatureFetcher(unleashConfig, this::handleStreamingUpdate);
     }
 
     public FeatureRepositoryImpl(
             UnleashConfig unleashConfig,
             BackupHandler featureBackupHandler,
             UnleashEngine engine,
-            FeatureFetcher fetcher) {
+            FeatureFetcher fetcher,
+            StreamingFeatureFetcher streamingFeatureFetcher) {
         this(
                 unleashConfig,
                 featureBackupHandler,
                 engine,
                 fetcher,
+                streamingFeatureFetcher,
                 unleashConfig.getToggleBootstrapProvider());
     }
 
@@ -61,12 +97,14 @@ public class FeatureRepositoryImpl implements FeatureRepository {
             BackupHandler featureBackupHandler,
             UnleashEngine engine,
             FeatureFetcher fetcher,
+            StreamingFeatureFetcher streamingFeatureFetcher,
             ToggleBootstrapProvider bootstrapHandler) {
         this(
                 unleashConfig,
                 featureBackupHandler,
                 engine,
                 fetcher,
+                streamingFeatureFetcher,
                 bootstrapHandler,
                 new EventDispatcher(unleashConfig));
     }
@@ -76,12 +114,14 @@ public class FeatureRepositoryImpl implements FeatureRepository {
             BackupHandler featureBackupHandler,
             UnleashEngine engine,
             FeatureFetcher fetcher,
+            StreamingFeatureFetcher streamingFeatureFetcher,
             ToggleBootstrapProvider bootstrapHandler,
             EventDispatcher eventDispatcher) {
         this.unleashConfig = unleashConfig;
         this.featureBackupHandler = featureBackupHandler;
         this.engine = engine;
         this.featureFetcher = fetcher;
+        this.streamingFeatureFetcher = streamingFeatureFetcher;
         this.bootstrapper = bootstrapHandler;
         this.eventDispatcher = eventDispatcher;
         this.throttler =
@@ -119,13 +159,17 @@ public class FeatureRepositoryImpl implements FeatureRepository {
             }
         }
 
-        if (!unleashConfig.isDisablePolling()) {
+        if (!unleashConfig.isDisablePolling() && !unleashConfig.isStreamingMode()) {
             Runnable updateFeatures = updateFeatures(this.eventDispatcher::dispatch);
             if (unleashConfig.getFetchTogglesInterval() > 0) {
                 executor.setInterval(updateFeatures, 0, unleashConfig.getFetchTogglesInterval());
             } else {
                 executor.scheduleOnce(updateFeatures);
             }
+        }
+
+        if (unleashConfig.isStreamingMode()) {
+            streamingFeatureFetcher.start();
         }
     }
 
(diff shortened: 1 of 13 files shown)


src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -201,4 +245,29 @@ public WasmResponse<VariantDef> getVariant(String toggleName, UnleashContext con
     public Stream<FeatureDefinition> listKnownToggles() {
         return this.engine.listKnownToggles().stream().map(FeatureDefinition::new);
     }
+
+    public void handleStreamingUpdate(String data) {
+        try {
+            engine.takeState(data);
+            // TODO: write backup when engine exposes current stage
```
kwasniew: yggdrasil needs to expose current state we can store to a backup file. Same problem we had in the Ruby SDK

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -201,4 +245,29 @@ public WasmResponse<VariantDef> getVariant(String toggleName, UnleashContext con
     public Stream<FeatureDefinition> listKnownToggles() {
         return this.engine.listKnownToggles().stream().map(FeatureDefinition::new);
     }
+
+    public void handleStreamingUpdate(String data) {
+        try {
+            engine.takeState(data);
```
kwasniew, line 242: ygg can already handle delta events

src/main/java/io/getunleash/repository/StreamingFeatureFetcherImpl.java
```diff
@@ -0,0 +1,148 @@
...
+            config.getCustomHttpHeadersProvider().getCustomHeaders().forEach(headersBuilder::add);
+
+            headersBuilder.add(UnleashConfig.UNLEASH_APP_NAME_HEADER, config.getAppName());
+            headersBuilder.add(UnleashConfig.UNLEASH_INSTANCE_ID_HEADER, config.getInstanceId());
+            headersBuilder.add(
+                    UnleashConfig.UNLEASH_CONNECTION_ID_HEADER, config.getConnectionId());
+            headersBuilder.add(UnleashConfig.UNLEASH_SDK_HEADER, config.getSdkVersion());
+            headersBuilder.add("Unleash-Client-Spec", config.getClientSpecificationVersion());
+
+            OkHttpClient httpClient =
```
kwasniew: in another PR I will tweak all SSE client settings across different SDKs. Each of the SSE clients uses different config options making it tricky

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -201,4 +231,35 @@ public WasmResponse<VariantDef> getVariant(String toggleName, UnleashContext con
     public Stream<FeatureDefinition> listKnownToggles() {
         return this.engine.listKnownToggles().stream().map(FeatureDefinition::new);
     }
+
+    public synchronized void handleStreamingUpdate(String data) {
```
kwasniew: decided to put this function here similar to updateFeatures for non streaming processing. The reson for that is that we have access to engine and dispatcher in current scope

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -0,0 +1,148 @@
...
+import java.net.URI;
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+import okhttp3.Headers;
+import okhttp3.OkHttpClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StreamingFeatureFetcherImpl implements StreamingFeatureFetcher {
```
kwasniew, line 17: This is the most important new file

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
gastonfournier: Thinking about this @sighphyre. I don't like "exposing internal state" of Yggdrasil. The SDK has no business with the internal state; it's just a proxy to the storage.

What if Ygg can write files? But that takes away the possibility of writing to other places (unless Ygg writes a file and returns a file handler, and the SDK does whatever with it). But what if instead of a file, Ygg returns a binary `byte[]`? Nobody needs to understand it, and it gives Ygg the flexibility to deal with the state. It can start by just sending a json, but later it can be something else.

Done that way, all SDKs will only need to know how to send that byte array to different storage locations

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
chriswk: Nice

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
chriswk: I guess the backup reader will have to check for another file, or we replace the existing backup of ClientFeatures with the backup of the EngineState, and trust Ygg to be able to read its own backup.

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -0,0 +1,148 @@
...
+
+        try {
+            BackgroundEventSource currentEventSource = eventSource;
+            if (currentEventSource != null) {
+                currentEventSource.close();
+                eventSource = null;
+            }
+            running.set(false);
+        } catch (Exception e) {
+            LOGGER.error("Error stopping streaming client", e);
```
gastonfournier:
```suggestion
            LOGGER.warn("Failed to stop streaming client", e);
```
I'd not consider this an error as this probably will only happen during shutdown, right?

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
kwasniew: I will pair with @sighphyre next week to think how to expose full internal flag state from ygg so that we can store it. Node SDK is doing exactly that and using existing backup mechanism (but not using Ygg yet).

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -0,0 +1,148 @@
...
+                            new UnleashEventHandler(), eventSourceBuilder);
+
+            BackgroundEventSource newEventSource = builder.build();
+            newEventSource.start();
+            eventSource = newEventSource;
+            running.set(true);
+        } catch (Exception e) {
+            LOGGER.error("Failed to start streaming client", e);
+            running.set(false);
+        }
```
gastonfournier: Here, there can be a race condition if 2 threads get the running state and proceed to start the stream.

I think you should do something like:
```java
try{
    if (!running.getAndSet(true)) {
        // start streaming
    } else {
        LOGGER.debug("Streaming client is already running");
    }
} catch (Exception e) {
    running.set(false);
}

**Review by chriswk — APPROVED** [review]
I like the granularity of the test here. Think Gaston's comments regarding race condition holds water.

A :+1: for me

**Review by gastonfournier — APPROVED** [review]
Looking great!

File: src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java (hunk 1 above)
Comment (sighphyre) [reply src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java]: > What if Ygg can write files?

I'd rather avoid this, right now Ygg does no IO outside of accessing randomness and time. I'd want a much stronger reason to change

> Ygg returns a binary byte[]? 

I was thinking just a string, honestly. That's the current interface into Ygg, seems fair to have be the interface out

Comment (gastonfournier) [reply src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java]: > I'd rather avoid this, right now Ygg does no IO outside of accessing randomness and time. I'd want a much stronger reason to change

+1

> I was thinking just a string, honestly

Yes, that'd work for me. String is almost == to byte[] if we agree on the encoding (which we did utf8)

File: src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java (hunk 4 above)
Comment (sighphyre) [inline src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java]: I think we can make these private. Might be nice to avoid a silly conversation in 2031 where a user has found this and has become dependent on it for some crazy reason

```suggestion
    private synchronized void handleStreamingUpdate(String data) {
```

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -201,4 +231,35 @@ public WasmResponse<VariantDef> getVariant(String toggleName, UnleashContext con
...
+            }
+        } catch (Exception e) {
+            LOGGER.error("Failed to process streaming update", e);
+            UnleashException unleashException =
+                    new UnleashException("Failed to process streaming update", e);
+            eventDispatcher.dispatch(unleashException);
+        }
+    }
+
+    public void handleStreamingError(Throwable error) {
```
sighphyre:
```suggestion
    private void handleStreamingError(Throwable error) {
```

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherFactory.java
```diff
@@ -0,0 +1,18 @@
+package io.getunleash.streaming;
```
sighphyre: I think this is just taste but since we only use this logic once do we need a class for it? I tend to prefer to have this logic close to where I need it so that it's clear what it's doing. I didn't guess this was the seam that fell back to the no-op impl when I saw this above

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    public void start() {
+        try {
+            if (!running.getAndSet(true)) {
+                // start streaming
+                URI streamingUri = config.getUnleashURLs().getStreamingURL().toURI();
+
+                Headers.Builder headersBuilder = new Headers.Builder();
+                config.getCustomHttpHeaders().forEach(headersBuilder::add);
```
sighphyre: This header setting code is kinda now repeated once here, once in UnleashConfig (for HttpFeatureFetcher) and once in OkHttpFeatureFetcher. Maybe doesn't have to be done in this PR but I think it's probably time that something exposes a getFetchHeaders function or something that can just return a map of all the headers we know need to get sent when grabbing features.

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -0,0 +1,150 @@
...
+            Consumer<String> streamingUpdateHandler,
+            Consumer<Throwable> streamingErrorHandler) {
+        this.config = config;
+        this.streamingUpdateHandler = streamingUpdateHandler;
+        this.streamingErrorHandler = streamingErrorHandler;
+    }
+
+    public void start() {
+        try {
+            if (!running.getAndSet(true)) {
```
sighphyre: I don't think we actually need the atomic boolean here. I peeked at the source of BackgroundEventSource and it's already doing this so we're just adding a second layer here. Unless you need to expose the isRunning method for something but that doesn't seem to be used anywhere

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        try {
+            BackgroundEventSource currentEventSource = eventSource;
+            if (currentEventSource != null) {
+                currentEventSource.close();
+                eventSource = null;
+            }
+            running.set(false);
+        } catch (Exception e) {
```
sighphyre, line 82: I went looking in the source to go find out how this fails if at all. I don't actually think it can. Still, better safe than sorry, good call

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -0,0 +1,150 @@
...
+            LOGGER.warn("Error stopping streaming client", e);
+        }
+    }
+
+    public boolean isRunning() {
```
sighphyre: Do we need this for something? I'd turf it if we don't. It's one less item in the public surface to cause problems

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -0,0 +1,150 @@
...
+                    case "unleash-connected":
+                    case "unleash-updated":
+                        streamingUpdateHandler.accept(messageEvent.getData());
+                        break;
+                    default:
+                        LOGGER.debug("Ignoring unknown event type: {}", event);
+                }
+
+            } catch (Exception e) {
+                LOGGER.error("Error processing streaming event: {}", event, e);
```
sighphyre: I'm a bit vague on the details around streaming but I believe this means it's game over for the SDK. We'll never be able to know if we have a consistently state of features until rehydration and the chances are we won't. Would something like this make that clearer to the end user that A Very Bad Thing™ has happened?
```suggestion
                LOGGER.error("Error processing streaming event, feature flags will likely not evaluate correctly until application restart: {}", event, e);
```

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherFactory.java
kwasniew: I'm glad you're saying that because that was my original version that I reluctantly extracted here lured by the smaller file size at the expense of local reasoning

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
kwasniew: You're 100% right here. Not I checked the source myself and it's doing the same thing internally.

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
kwasniew: not anymore. leftover from a previous version

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
kwasniew: Yeah, good idea. Only application restart or re-connect will recover the situation

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
kwasniew: Yeah, I'd prefer to add some duplication in this PR and clean up in a separate PR

PR #322, opened 2025-08-20
feat: streaming example

kwasniew:
## About the changes

Closes #


examples/streaming-example/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+    mainClass.set("io.getunleash.example.StreamingExample")
+}
+
+repositories {
+    mavenCentral()
+    mavenLocal()
+}
+
+dependencies {
+    implementation("io.getunleash:unleash-client-java:11.1.0")
```
kwasniew, line 16: version with streaming required

examples/streaming-example/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+    implementation("com.launchdarkly:okhttp-eventsource:4.1.1")
```
kwasniew, line 17: optional dependency for streaming

gastonfournier:
I manually tested it against a server without streaming support, and it doesn't fall back to polling. I don't know if it's the intended behavior but it does use the backup to load the initial state but it does not fetch from upstream.

examples/streaming-example/src/main/java/io/getunleash/example/StreamingExample.java
```diff
@@ -0,0 +1,65 @@
...
+
+        System.out.println("Streaming client started. Waiting for events... (Press Ctrl+C to exit)");
+
+        try {
+            Thread.currentThread().join();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            System.out.println("Streaming example shutting down");
+        }
+    }
```
gastonfournier, line 56: Looks ok, I'd like to also get some feature evaluation result, but I tested that manually and it works well.

I did something similar to this: https://github.com/Unleash/unleash-java-sdk/blob/41c6028287f707e1eac3d77d69eeb00f1039860a/examples/cli-example/src/main/java/io/getunleash/example/AdvancedConstraints.java#L29-L32 but limiting to 100 iterations with a sleep.

gastonfournier (approved):
This looks good! The happy path works well. We know about this limitation https://github.com/Unleash/unleash-java-sdk/pull/322#issuecomment-3258233520 but it will be addressed later. Meanwhile, the features is clearly marked as experimental which means, it should not be used in production yet. Good step forward!

examples/streaming-example/src/main/java/io/getunleash/example/StreamingExample.java
kwasniew: There's `boolean isEnabled = unleash.isEnabled("streaming_flag");` which is a flag you need to create and you will see it changing instantly on every streaming change. This is a similar pattern I did in Node SDK where we react to instant changes instead of looping which was fine for the polling mindset but for streaming event listening mindset seems to be more appropriate for me.

PR #333, opened 2025-11-10
feat: change engine proxy to use flatbuffer implementation over wasm

chriswk:
This patches v11, we should probably also look at updating v10. But for now, this PR allows us to create an RC/beta that can run in our infrastructure to allow us to verify memory usage and performance.


src/main/java/io/getunleash/repository/FeatureRepository.java
```diff
@@ -2,15 +2,17 @@
...
 import java.util.stream.Stream;
 
 public interface FeatureRepository {
 
-    WasmResponse<Boolean> isEnabled(String toggleName, UnleashContext context);
+    @Nullable
+    FlatResponse<Boolean> isEnabled(String toggleName, UnleashContext context);
 
-    WasmResponse<VariantDef> getVariant(String toggleName, UnleashContext context);
+    FlatResponse<VariantDef> getVariant(String toggleName, UnleashContext context);
```
sjaanus, line 16: This seems also nullable.

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -145,7 +143,7 @@ public Variant getVariant(String toggleName, Variant defaultValue) {
     @Override
     public Variant getVariant(String toggleName, UnleashContext context, Variant defaultValue) {
         UnleashContext enhancedContext = context.applyStaticFields(config);
-        WasmResponse<VariantDef> response =
+        FlatResponse<VariantDef> response =
                 this.featureRepository.getVariant(toggleName, enhancedContext);
         Optional<VariantDef> variantDef = Optional.ofNullable(response.value);
```
sjaanus: This can give nullpointer exception, if response is null.

src/main/java/io/getunleash/repository/FeatureRepository.java
chriswk: yeah, true, thanks

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -145,13 +143,13 @@ public Variant getVariant(String toggleName, Variant defaultValue) {
     @Override
     public Variant getVariant(String toggleName, UnleashContext context, Variant defaultValue) {
         UnleashContext enhancedContext = context.applyStaticFields(config);
-        WasmResponse<VariantDef> response =
-                this.featureRepository.getVariant(toggleName, enhancedContext);
-        Optional<VariantDef> variantDef = Optional.ofNullable(response.value);
+        Optional<FlatResponse<VariantDef>> response =
+                Optional.ofNullable(this.featureRepository.getVariant(toggleName, enhancedContext));
+        Optional<VariantDef> variantDef = response.map(r -> r.value);
```
sjaanus, line 148: I think now if we call `variantDef.get()` you might get null as result, but you actually want `isPresent` to return false?

I think we want `Optional<VariantDef> variantDef = response.flatMap(r -> Optional.ofNullable(r.value));`

This is what we do in https://github.com/Unleash/unleash-java-sdk/pull/333/files#diff-5209f3f32d148f4e260e39b0166a5cd032fdc210bd83b0859993c254d4325263R119

chriswk: I agree monadically / if java actually did optionals the expected way, but as I just confirmed with jshell, when you map an option, if your option returns null, you get Optional.empty, not an NPE

```
jshell> Optional<String> my_present = Optional.of("hello");
my_present ==> Optional[hello]

jshell> Optional<String> absent = my_present.map(s -> null)
absent ==> Optional.empty
```

sjaanus: Oh wow, nice!

examples/cli-example/src/main/java/io/getunleash/example/GetVariant.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class GetVariant {
+    public static void main(String[] args) throws InterruptedException {
+        UnleashConfig config = UnleashConfig.builder()
+            .appName("client-example.advanced.java")
+            .customHttpHeader(
+                "Authorization",
+                getOrElse("UNLEASH_API_TOKEN",
+                    "*:development.25a06b75248528f8ca93ce179dcdd141aedfb632231e0d21fd8ff349"))
+            .unleashAPI(getOrElse("UNLEASH_API_URL", "https://app.unleash-hosted.com/demo/api"))
```
daveleek, line 18: Good enough for now I guess, but yeah maybe should've been env arg?

examples/cli-example/src/main/java/io/getunleash/example/KnownToggles.java
```diff
@@ -0,0 +1,35 @@
...
+import static io.getunleash.example.AdvancedConstraints.getOrElse;
+
+public class KnownToggles {
+    public static void main(String[] args) throws InterruptedException {
+        UnleashConfig config = UnleashConfig.builder()
+            .appName("client-example.advanced.java")
+            .customHttpHeader(
+                "Authorization",
+                getOrElse("UNLEASH_API_TOKEN",
+                    "*:development.25a06b75248528f8ca93ce179dcdd141aedfb632231e0d21fd8ff349"))
```
daveleek, line 17: Same

examples/spring-boot-example/src/main/resources/application.yaml
```diff
@@ -1,4 +1,8 @@
 unleash:
-    url: https://app.unleash-hosted.com/demo/api
-    apiKey: demo-app:production.614a75cf68bef8703aa1bd8304938a81ec871f86ea40c975468eabd6
+    url: http://localhost:4242/api
+    apiKey: "*:development.da33883b2b2be02c5648ac624a1f50922cbc5893c2e6de5b9f57298d"
```
daveleek, line 3: this isn't going to work is it?

src/main/java/io/getunleash/repository/FeatureRepository.java
daveleek: .NET is expecting these to not be null I believe

daveleek (approved):
Cool :)

examples/cli-example/src/main/java/io/getunleash/example/GetVariant.java
chriswk: It is an env arg. we just default to one that used to be able to access demo, but since we wipe demo regularly, I'm not worried.

examples/cli-example/src/main/java/io/getunleash/example/KnownToggles.java
chriswk: same

PR #334, opened 2025-12-04
chore: resolve a bunch of linter warnings, no logic changes

sighphyre:
Don't think this makes any logical changes but it stops my IDE screaming at me in a bunch of places

### DIFF
--- a/src/main/java/io/getunleash/DefaultUnleash.java
+++ b/src/main/java/io/getunleash/DefaultUnleash.java
@@ -8,7 +8,6 @@ import io.getunleash.event.EventDispatcher;
 import io.getunleash.event.IsEnabledImpressionEvent;
 import io.getunleash.event.ToggleEvaluated;
 import io.getunleash.event.VariantImpressionEvent;
-import io.getunleash.metric.UnleashMetricService;
 import io.getunleash.repository.FeatureRepository;
 import io.getunleash.repository.YggdrasilAdapters;
 import io.getunleash.strategy.*;
@@ -27,7 +26,6 @@ public class DefaultUnleash implements Unleash {
 
     private static ConcurrentHashMap<String, LongAdder> initCounts = new ConcurrentHashMap<>();
 
-    private final UnleashMetricService metricService;
     private final FeatureRepository featureRepository;
     private final UnleashContextProvider contextProvider;
     private final EventDispatcher eventDispatcher;
@@ -69,7 +67,6 @@ public class DefaultUnleash implements Unleash {
 
         this.config = unleashConfig;
         this.featureRepository = engineProxy;
-        this.metricService = engineProxy;
         this.contextProvider = contextProvider;
         this.eventDispatcher = eventDispatcher;
         initCounts.compute(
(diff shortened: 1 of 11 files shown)


src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -27,7 +26,6 @@ public class DefaultUnleash implements Unleash {
 
     private static ConcurrentHashMap<String, LongAdder> initCounts = new ConcurrentHashMap<>();
 
-    private final UnleashMetricService metricService;
```
sighphyre, line 30: All handled by the EngineProxy

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
```diff
@@ -8,11 +8,8 @@
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.Set;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class UnleashMetricServiceImpl implements UnleashMetricService {
-    private static final Logger LOGGER = LoggerFactory.getLogger(UnleashMetricServiceImpl.class);
```
sighphyre, line 15: Not used

src/main/java/io/getunleash/util/UnleashScheduledExecutorImpl.java
```diff
@@ -48,7 +48,11 @@ public void setInterval(Runnable command, long initialDelaySec, long periodSec)
 
     @Override
     public Future<Void> scheduleOnce(Runnable runnable) {
-        return (Future<Void>) executorService.submit(runnable);
+        return executorService.submit(
```
sighphyre, line 51: Think this was just wrong but didn't matter because the result wasn't really used. Pretty sure this is correct but still won't matter

Linter stops freaking though, which is nice

src/test/java/io/getunleash/DefaultUnleashTest.java
```diff
@@ -23,9 +23,7 @@
 import io.getunleash.util.UnleashConfig;
```
sighphyre, line 23: Mostly a bunch of unuseds which don't impact the test runs in anyway

src/test/java/io/getunleash/UnleashTest.java
```diff
@@ -59,6 +59,7 @@ public void fallback_function_should_be_invoked_and_return_true() {
                 .thenReturn(new FlatResponse<Boolean>(false, null));
         Unleash unleash = new DefaultUnleash(baseConfig, engineProxy);
 
+        @SuppressWarnings("unchecked")
```
sighphyre, line 62: Think this is correct thing to do here. Mocking is runtime goop so I don't think this has a meaningful way to determine the type

src/test/java/io/getunleash/streaming/StreamingFeatureFetchingTest.java
```diff
@@ -1,4 +1,4 @@
-package io.getunleash.integration;
+package io.getunleash.streaming;
```
sighphyre, line 1: Not sure why this was in the integration package. It's alone in there, doesn't match it's actual package structure and that doesn't follow the pattern for our test packages anyway

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -69,7 +67,6 @@ public DefaultUnleash(
 
         this.config = unleashConfig;
         this.featureRepository = engineProxy;
-        this.metricService = engineProxy;
```
daveleek, line 72: > All handled by the EngineProxy

But but..

PR #336, opened 2025-12-15
feat: impact metric counter

sjaanus:
Add first iteration of impact metrics. Counters. Next up is gauges.

### DIFF
--- a/src/test/java/io/getunleash/impactmetrics/InMemoryMetricRegistryTest.java
+++ b/src/test/java/io/getunleash/impactmetrics/InMemoryMetricRegistryTest.java
@@ -12,7 +12,7 @@ public class InMemoryMetricRegistryTest {
     @Test
     public void should_increment_by_default_value() {
         InMemoryMetricRegistry registry = new InMemoryMetricRegistry();
-        Counter counter = registry.counter(new MetricOptions("test_counter", "test help"));
+        Counter counter = registry.counter(new MetricOptions("test_counter", "testing"));
 
         counter.inc();
 
@@ -20,7 +20,7 @@ public class InMemoryMetricRegistryTest {
 
         CollectedMetric expected =
                 new CollectedMetric(
-                        "test_counter", "test help", MetricType.COUNTER, List.of(sample(1L)));
+                        "test_counter", "testing", MetricType.COUNTER, List.of(sample(1L)));
 
         assertThat(metrics).containsExactly(expected);
     }


src/main/java/io/getunleash/impactmetrics/CounterImpl.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+        ConcurrentHashMap<String, String> labels = new ConcurrentHashMap<>();
+        for (String pair : key.split(",")) {
+            String[] parts = pair.split("=", 2);
+            if (parts.length == 2) {
+                labels.put(parts[0], parts[1]);
+            }
+        }
+        return labels;
+    }
```
sjaanus, line 76: Following same logic as we have in node SDK.

src/main/java/io/getunleash/impactmetrics/CounterImpl.java
```diff
@@ -0,0 +1,75 @@
...
+
+        values.forEach(
+                (key, value) -> {
+                    samples.add(new NumericMetricSample(parseLabelKey(key), value));
+                });
+        values.clear();
+
+        if (samples.isEmpty()) {
+            samples.add(new NumericMetricSample(Collections.emptyMap(), 0L));
+        }
```
sjaanus, line 49: Populate with default value to at least have something.

src/main/java/io/getunleash/impactmetrics/MetricType.java
```diff
@@ -0,0 +1,7 @@
+package io.getunleash.impactmetrics;
+
+public enum MetricType {
+    COUNTER,
+    GAUGE,
+    HISTOGRAM
+}
```
sjaanus, line 7: Preparing for gauge and histogram.

src/main/java/io/getunleash/impactmetrics/CollectedMetric.java
```diff
@@ -0,0 +1,34 @@
+package io.getunleash.impactmetrics;
+
+import java.util.List;
+
+public class CollectedMetric {
+    private final String name;
+    private final String help;
+    private final MetricType type;
+    private final List<NumericMetricSample> samples;
```
kwasniew, line 10: this will nor work for histograms

kwasniew: they have BucketMetricSamples

src/main/java/io/getunleash/impactmetrics/CounterImpl.java
```diff
@@ -0,0 +1,75 @@
...
+                .sorted(Map.Entry.comparingByKey())
+                .map(e -> e.getKey() + "=" + e.getValue())
+                .collect(Collectors.joining(","));
+    }
+
+    static Map<String, String> parseLabelKey(String key) {
+        if (key == null || key.isEmpty()) {
+            return Collections.emptyMap();
+        }
+        ConcurrentHashMap<String, String> labels = new ConcurrentHashMap<>();
```
kwasniew: you're not sharing this so Map<String, String> labels = new HashMap<>(); would suffice.

src/main/java/io/getunleash/impactmetrics/CounterImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public CollectedMetric collect() {
+        List<NumericMetricSample> samples = new ArrayList<>();
+
+        values.forEach(
+                (key, value) -> {
+                    samples.add(new NumericMetricSample(parseLabelKey(key), value));
+                });
+        values.clear();
```
kwasniew: increments between forEach and clear can be lost. Java is not running single threaded event loop with run to completion guarantee as Node

src/main/java/io/getunleash/impactmetrics/CollectedMetric.java
sjaanus: Yes histograms is different, but I will solve it when I get there.

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Test
+    public void should_increment_by_default_value() {
+        CounterImpl counter = new CounterImpl("test_counter", "test help");
+
+        counter.inc();
+
+        CollectedMetric result = counter.collect();
+
+        assertThat(result.getName()).isEqualTo("test_counter");
```
kwasniew: looks like missing equals implementation on a result

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+public class CounterImplTest {
+
+    @Test
+    public void should_increment_by_default_value() {
+        CounterImpl counter = new CounterImpl("test_counter", "test help");
```
kwasniew: it doesn't follow testing convention from node where we have registry, counter on a registry and registry verification

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        counter.inc(1, Map.of("a", "x"));
+        counter.inc(2, Map.of("b", "y"));
+        counter.inc(3);
+
+        CollectedMetric result = counter.collect();
+
+        assertThat(result.getSamples()).hasSize(3);
+
+        NumericMetricSample sampleAX =
```
kwasniew: this logic it too complex for unit tests that should be declarative. we should strive for a single assertion as in node. prbbaly another missing equals implementation leaking complexity to the unit test

src/main/java/io/getunleash/impactmetrics/CounterImpl.java
sjaanus: good point, fixing.

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        assertThat(key1).isEqualTo(key2);
+        assertThat(key1).isEqualTo("a=2,z=1");
+    }
+
+    @Test
+    public void should_parse_label_key_correctly() {
+        String key = "env=prod,region=us-east";
+
+        Map<String, String> labels = CounterImpl.parseLabelKey(key);
```
kwasniew: this is private bahavior that should not be exposed and only tested transitively as part of the core behavior

kwasniew: Node doesn't export parseLabelKey

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
sjaanus: The point was to check that things do not get bucketed in wrong places and provide real world example, with different buckets.

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+        assertThat(result.getName()).isEqualTo("test_counter");
+        assertThat(result.getHelp()).isEqualTo("test help");
+        assertThat(result.getType()).isEqualTo(MetricType.COUNTER);
+        assertThat(result.getSamples()).hasSize(1);
+        assertThat(result.getSamples().get(0).getValue()).isEqualTo(1L);
+        assertThat(result.getSamples().get(0).getLabels()).isEmpty();
+    }
+
+    @Test
+    public void should_increment_with_custom_value() {
```
kwasniew: in Node we had one test for values and labels. keeping test structure similar make it easier to review them side by side

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+        assertThat(second.getSamples()).hasSize(1);
+        assertThat(second.getSamples().get(0).getValue()).isEqualTo(0L);
+    }
+
+    @Test
+    public void should_generate_stable_label_key() {
+        Map<String, String> labels1 = Map.of("z", "1", "a", "2");
+        Map<String, String> labels2 = Map.of("a", "2", "z", "1");
+
+        String key1 = CounterImpl.getLabelKey(labels1);
```
kwasniew: same as the next comment

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
sjaanus: we do not have registry yet.

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
kwasniew: Sure, I agree with the intent. My point is that we're still missing equals and hashCode on NumericMetricSample and CollectedMetric that would make some of the checks easier to parse

kwasniew: what we do as toMatchObject in Node for value object equality checks is typically implemented in equals methods in Java and tests don't need to dissect every single field in a separate assertion

kwasniew: e.g. 
```
        assertThat(result.getSamples())
                .containsExactlyInAnyOrder(
                        sample(Map.of("a", "x"), 1L), sample(Map.of("b", "y"), 2L), sample(3L))
```

src/main/java/io/getunleash/impactmetrics/CounterImpl.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+public class CounterImpl implements Counter {
```
kwasniew: if we want to keep it compliant with Node then it should not be public

src/test/java/io/getunleash/impactmetrics/CounterImplTest.java
kwasniew: the test is showing behavior we don't want to expose as public. I think Node deliberately does not export CounterImpl

kwasniew (approved):

PR #339, opened 2025-12-17
feat: impact metrics histogram

sjaanus:
- Added histogram metric support
- Introduced bucket-based metric options
- Added histogram unit test coverage

### DIFF
--- a/src/test/java/io/getunleash/impactmetrics/MetricTypeTest.java
+++ b/src/test/java/io/getunleash/impactmetrics/MetricTypeTest.java
@@ -7,7 +7,7 @@ import java.util.List;
 import java.util.Map;
 import org.junit.jupiter.api.Test;
 
-public class InMemoryMetricRegistryTest {
+public class MetricTypeTest {
 
     @Test
     public void should_increment_by_default_value() {


src/main/java/io/getunleash/impactmetrics/HistogramImpl.java
```diff
@@ -0,0 +1,104 @@
...
+            inputBuckets = DEFAULT_BUCKETS;
+        }
+
+        List<Double> sorted =
+                inputBuckets.stream()
+                        .filter(d -> d != Double.POSITIVE_INFINITY)
+                        .distinct()
+                        .sorted()
+                        .collect(Collectors.toList());
+        sorted.add(Double.POSITIVE_INFINITY);
```
sjaanus, line 31: In java there is no such thing as `+Inf` as string and hard to model. In java there is `POSITIVE_INFINITY` and we can later change to `+Inf` if need to send to Unleash.

src/main/java/io/getunleash/impactmetrics/CollectedMetric.java
```diff
@@ -1,20 +1,21 @@
...
 public class CollectedMetric {
     private final String name;
     private final String help;
     private final MetricType type;
-    private final List<NumericMetricSample> samples;
+    private final List<MetricSample> samples;
 
     public CollectedMetric(
-            String name, String help, MetricType type, List<NumericMetricSample> samples) {
+            String name, String help, MetricType type, List<? extends MetricSample> samples) {
```
sjaanus, line 14: We now have 2 types that extend `MetricSample`.

src/main/java/io/getunleash/impactmetrics/HistogramImpl.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public void observe(double value, Map<String, String> labels) {
+        String key = CounterImpl.getLabelKey(labels);
+        values.compute(
+                key,
+                (k, data) -> {
+                    if (data == null) {
+                        data = new HistogramData(0, 0.0, buckets);
+                    }
+                    synchronized (data) {
```
sjaanus, line 49: We need to use synchronized due to concurrency issues. We might have multiple concurrent calls to observe and it is manipulating the HistogramData object.

src/main/java/io/getunleash/impactmetrics/HistogramImpl.java
kwasniew: how do you send it over the wire in serialized JSON?

sjaanus: This is next step. Probably just transform it before sending.

src/test/java/io/getunleash/impactmetrics/MetricTypeTest.java
kwasniew, line 1: why did you decide to change a test file name compared to Node SDK? I'd stay to MetricTypeTest.java I think it's ok that we don't call tests after classes since we're testing behaviors not classes

kwasniew (approved):

PR #341, opened 2025-12-18
feat: start sending impact metrics in payload

sjaanus:
Start sending impact metrics in metrics payload.

### DIFF
--- a/src/main/java/io/getunleash/util/UnleashConfig.java
+++ b/src/main/java/io/getunleash/util/UnleashConfig.java
@@ -7,6 +7,7 @@ import io.getunleash.UnleashException;
 import io.getunleash.event.NoOpSubscriber;
 import io.getunleash.event.UnleashSubscriber;
 import io.getunleash.impactmetrics.ImpactMetricsDataSource;
+import io.getunleash.impactmetrics.InMemoryMetricRegistry;
 import io.getunleash.lang.Nullable;
 import io.getunleash.metric.DefaultHttpMetricsSender;
 import io.getunleash.repository.HttpFeatureFetcher;
@@ -766,7 +767,7 @@ public class UnleashConfig {
                     proxyAuthenticator,
                     startupExceptionHandler,
                     Optional.ofNullable(impactMetricsRegistry)
-                            .orElseGet(io.getunleash.impactmetrics.InMemoryMetricRegistry::new));
+                            .orElseGet(InMemoryMetricRegistry::new));
         }
 
         public String getDefaultSdkVersion() {


src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
```diff
@@ -59,13 +68,21 @@ private Runnable sendMetrics() {
             if (throttler.performAction()) {
                 MetricsBucket bucket = this.engine.getMetrics();
 
-                ClientMetrics metrics = new ClientMetrics(unleashConfig, bucket);
+                List<CollectedMetric> impactMetrics = impactMetricsRegistry.collect();
+                List<CollectedMetric> impactMetricsOrNull =
+                        impactMetrics.isEmpty() ? null : impactMetrics;
+
+                ClientMetrics metrics =
+                        new ClientMetrics(unleashConfig, bucket, impactMetricsOrNull);
```
sjaanus, line 73: There is design from node SDK, that if we do not have any metrics, we do not add the empty array to payload.

src/test/java/io/getunleash/impactmetrics/MetricTypeTest.java
```diff
@@ -303,60 +300,4 @@ private NumericMetricSample sample(long value) {
     private NumericMetricSample sample(Map<String, String> labels, long value) {
         return new NumericMetricSample(labels, value);
     }
-
-    @Test
-    public void should_serialize_histogram_with_positive_infinity_as_plus_inf_string() {
```
sjaanus, line 308: Moving this to http level, so we can catch also if serializer was not mounted.

src/test/java/io/getunleash/metric/DefaultHttpMetricsSenderTest.java
```diff
@@ -108,4 +113,44 @@ public void should_handle_service_failure_when_sending_metrics() throws URISynta
                         .withHeader(UNLEASH_INTERVAL, matching(String.valueOf(metricsInterval)))
                         .withHeader("UNLEASH-APPNAME", matching("test-app")));
     }
+
+    @Test
+    public void should_send_impact_metrics_with_histogram_and_plus_inf_bucket()
```
sjaanus, line 118: Following the java test style, we verify that `+Inf `is being sent.

src/test/java/io/getunleash/metric/UnleashMetricServiceImplTest.java
```diff
@@ -359,6 +363,129 @@ public void url_not_found_immediately_increases_interval_to_max() throws Yggdras
         assertThat(unleashMetricService.getSkips()).isEqualTo(0);
     }
 
+    @Test
+    public void should_include_impact_metrics_in_clientmetrics_payload() throws YggdrasilError {
```
sjaanus, line 367: These tests are same as in node SDK.

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
```diff
@@ -42,6 +48,9 @@ public UnleashMetricServiceImpl(
                         300,
                         unleashConfig.getUnleashURLs().getClientMetricsURL());
         this.engine = engine;
+        ImpactMetricsDataSource configuredRegistry = unleashConfig.getImpactMetricsRegistry();
```
kwasniew: why the registry lives in the config and is not passed as other dependencies via constructor?

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
```diff
@@ -42,6 +48,9 @@ public UnleashMetricServiceImpl(
...
+        this.impactMetricsRegistry =
```
kwasniew: this is a code smell (https://lab.abilian.com/Tech/Architecture%20%26%20Software%20Design/Dependency%20Inversion/DI%20anti-patterns/#bastard-injection-poor-mans-injection). as a user I don't know when I should pass the registry and when it will be created automatically.

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
sjaanus: `ImpactMetricsDataSource` is configurable, so it belongs in `UnleashConfig`, and `UnleashMetricServiceImpl` reads it from there (with a default fallback), just like it does for other configurable dependencies.

src/main/java/io/getunleash/metric/UnleashMetricServiceImpl.java
sjaanus: Good catch, fixed it by following what we do with `UnleashScheduledExecutorImpl`

kwasniew (approved):

PR #347, opened 2026-01-05
chore: failover strategy

sighphyre:
Adds an unwired up failover strategy. The intent here is to use this in the streaming event handling to determine whether or not a given error event should cause a failover to polling. Currently this does nothing useful except determine whether or not a failover should occur. Next PR will add actual failover

### DIFF
--- a/src/test/java/io/getunleash/streaming/FailoverStrategyTest.java
+++ b/src/test/java/io/getunleash/streaming/FailoverStrategyTest.java
@@ -15,7 +15,7 @@ class FailoverStrategyTest {
         FailoverStrategy strategy = new FailoverStrategy(3, 5_000);
         ServerEvent event = new ServerEvent(Instant.now(), "Switch to polling", "polling");
 
-        assertThat(strategy.shouldFailover(event)).isTrue();
+        assertThat(strategy.shouldFailover(event, Instant.now())).isTrue();
     }
 
     @Test
@@ -23,7 +23,7 @@ class FailoverStrategyTest {
         FailoverStrategy strategy = new FailoverStrategy(3, 5_000);
         ServerEvent event = new ServerEvent(Instant.now(), "Restart complete", "unleash-restarted");
 
-        assertThat(strategy.shouldFailover(event)).isFalse();
+        assertThat(strategy.shouldFailover(event, Instant.now())).isFalse();
     }
 
     @Test
@@ -31,7 +31,7 @@ class FailoverStrategyTest {
         FailoverStrategy strategy = new FailoverStrategy(3, 5_000);
         HttpStatusError error = new HttpStatusError(Instant.now(), "Too many connections", 429);
 
-        assertThat(strategy.shouldFailover(error)).isTrue();
+        assertThat(strategy.shouldFailover(error, Instant.now())).isTrue();
     }
 
     @Test


src/main/java/io/getunleash/streaming/FailoverStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+
+        switch (event.getType()) {
+            case HTTP_STATUS_ERROR:
+                return handleHttpStatus((HttpStatusError) event);
+            case SERVER_HINT:
+                return handleServerEvent((ServerEvent) event);
+            case NETWORK_ERROR:
+                return handleNetwork(event);
+            default:
+                LOGGER.warn(
```
sighphyre, line 40: This should never happen, unless it does. If it does happen it's because someone extended the enum without handling it here. In Rust/TS we can make this a compile error :) In Java we can't :(

src/main/java/io/getunleash/streaming/FailoverStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+        return false;
+    }
+
+    private boolean hasTooManyFails(FailEvent event) {
+        failures.add(event);
+        return failures.size() >= maxFails;
+    }
+
+    private void pruneOldFailures(long nowMs) {
```
sighphyre, line 69: In place list compaction with no extra allocations. I dunno if this is clear enough, I think it's okay. Happy to change it if anyone hates it

src/main/java/io/getunleash/streaming/FailoverStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class FailoverStrategy {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FailoverStrategy.class);
+    private static final Set<String> FAILOVER_SERVER_HINTS = Set.of("polling");
+    private static final Set<Integer> HARD_FAILOVER_STATUS_CODES = Set.of(401, 403, 404, 429, 501);
+    private static final Set<Integer> SOFT_FAILOVER_STATUS_CODES = Set.of(408, 500, 502, 503, 504);
```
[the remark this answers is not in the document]
sighphyre: Yes I hear you but in this case you're wrong. The most likely cause here is that upstream isn't exposing it. In this case, the SDK needs to failover now, otherwise toggles will not evaluate correctly

src/main/java/io/getunleash/streaming/FailoverStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private final int maxFails;
+    private final long relaxTimeMs;
+    private final List<FailEvent> failures = new ArrayList<>();
```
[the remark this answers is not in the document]
sighphyre: I'm not actually sure this is true but sync costs very little here since it's a cold path so added

src/main/java/io/getunleash/streaming/FailoverStrategy.java
```diff
@@ -0,0 +1,167 @@
...
+    private boolean handleServerEvent(ServerEvent event) {
+        return FAILOVER_SERVER_HINTS.contains(event.getEvent());
+    }
+
+    private boolean handleNetwork(FailEvent event) {
+        return hasTooManyFails(event);
+    }
+
+    private boolean handleHttpStatus(HttpStatusError event) {
+        if (HARD_FAILOVER_STATUS_CODES.contains(event.getStatusCode())) {
```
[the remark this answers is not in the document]
sighphyre: Not a bug. Intentional. Doesn't need to be recorded

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -116,6 +136,61 @@ synchronized void handleStreamingUpdate(String data) {
         }
     }
 
+    void handleStreamingError(Throwable throwable) {
+        handleFailoverDecision(toFailEvent(throwable));
+    }
+
+    void handleModeChange(String eventData) {
+        if (eventData.equals("polling")) {
```
sighphyre, line 144: Magic string but used once so not sure I care to pull it into a constant

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
```diff
@@ -116,6 +136,61 @@ synchronized void handleStreamingUpdate(String data) {
...
+    }
+
+    private void handleFailoverDecision(FailoverStrategy.FailEvent failEvent) {
+        boolean shouldFail = failoverStrategy.shouldFailover(failEvent, Instant.now());
+        if (shouldFail) {
+            LOGGER.warn(
+                    "Streaming failover triggered: {}. Client is switching over to polling mode.",
+                    failEvent.getMessage());
+
+            // changeToPolling()
```
sighphyre, line 190: Incoming in the next PR, this is where the hook lies to trigger a failover

src/main/java/io/getunleash/streaming/StreamingFeatureFetcherImpl.java
chriswk: I'm fine with the first two to three of these... :)

src/test/java/io/getunleash/streaming/FailoverStrategyTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertThat(strategy.shouldFailover(event, Instant.now())).isFalse();
+    }
+
+    @Test
+    void should_fail_immediately_on_hard_http_status() {
+        FailoverStrategy strategy = new FailoverStrategy(3, 5_000);
+        HttpStatusError error = new HttpStatusError(Instant.now(), "Too many connections", 429);
+
+        assertThat(strategy.shouldFailover(error, Instant.now())).isTrue();
+    }
```
chriswk, line 35: I guess it is better to poll than stream if we're overloaded

sighphyre: Yeah it's explicit handling code on the Unleash side - we cap streaming connections to a set amount, any incoming request that would go over that limit gets a 429, so it either needs to poll or die. Polling sounds like a nicer option haha

PR #348, opened 2026-01-08
chore: wraps streaming/polling in a class that can make intelligent decisions about which to use

sighphyre:
Adds an AdaptiveFetcher ala the Node SDK, which allows us to make intelligent decisions around whether or not to use streaming. Done a bit of clean up here around packages and tests too

### DIFF
--- a/src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
+++ b/src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
@@ -52,7 +52,7 @@ public class FeatureRepositoryImpl implements FeatureRepository {
 
         this.featureBackupHandler = featureBackupHandler;
         this.engine = engine;
-        this.bootstrapper = unleashConfig.getToggleBootstrapProvider();
+        this.bootstrapper = bootstrapHandler;
         this.eventDispatcher = readyOnceGate;
         this.fetcher = fetcher;
         this.initCollections(unleashConfig.getScheduledExecutor());


src/main/java/io/getunleash/repository/AdaptiveFetcher.java
```diff
@@ -0,0 +1,89 @@
+package io.getunleash.repository;
+
+import io.getunleash.engine.UnleashEngine;
+import io.getunleash.event.EventDispatcher;
+import io.getunleash.event.GatedEventEmitter;
+import io.getunleash.util.UnleashConfig;
+
+public class AdaptiveFetcher implements FetchWorker, ModeController {
+
+    interface WorkersProvider {
```
sighphyre, line 10: There's two way communication between the adaptive fetcher and it's kids because the kids need to signal a failover event and this is the thing that wraps them. So we do this dance where we pass in an interface that can construct the workers, this gives us late binding, otherwise we have a chicken and egg scenario

src/main/java/io/getunleash/repository/FailoverStrategy.java
```diff
@@ -1,6 +1,6 @@
 // reference implementation here: https://github.com/Unleash/unleash-node-sdk/pull/780
 
-package io.getunleash.streaming;
+package io.getunleash.repository;
```
sighphyre, line 3: I've collapsed the streaming package back into the repository package. In theory I love the split here but in practice it means that a domain is split over two packages, which requires that the contained classes are public. Which is a worse problem than just having a fat package

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -36,23 +33,10 @@ public FeatureRepositoryImpl(UnleashConfig unleashConfig, UnleashEngine engine)
...
-                        engine,
-                        featureBackupHandler,
-                        readyOnceGate);
-        this.streamingFeatureFetcher =
-                new StreamingFeatureFetcherImpl(
-                        unleashConfig,
-                        new EventDispatcher(unleashConfig),
-                        engine,
-                        featureBackupHandler);
+        this.fetcher = new AdaptiveFetcher(unleashConfig, featureBackupHandler, engine);
```
sighphyre, line 39: We now depend on a single fetch worker here. By default this is the adaptive fetcher but in theory you can and should be able to inject anything that conforms to the fetching interface and it should still work. Which is nice, I think

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
```diff
@@ -61,53 +45,16 @@ public FeatureRepositoryImpl(UnleashConfig unleashConfig, UnleashEngine engine)
             UnleashConfig unleashConfig,
             BackupHandler featureBackupHandler,
             UnleashEngine engine,
-            FeatureFetcher fetcher,
-            FetchWorker streamingFeatureFetcher) {
```
sighphyre, line 65: Yeeting a bunch of constructors. These are maintained for tests, which is daft. We can just construct things in setup functions in tests if we need complex setup

src/main/java/io/getunleash/repository/AdaptiveFetcher.java
```diff
@@ -0,0 +1,89 @@
+package io.getunleash.repository;
+
+import io.getunleash.engine.UnleashEngine;
+import io.getunleash.event.EventDispatcher;
+import io.getunleash.event.GatedEventEmitter;
+import io.getunleash.util.UnleashConfig;
+
+public class AdaptiveFetcher implements FetchWorker, ModeController {
```
sighphyre, line 8: The adaptive fetcher is a fetchworker, which is a thing that's responsible for getting flags, but it's also a mode controller, which another entity can make a request that it fails over to (in this case, the streaming fetcher)

gastonfournier: This class assumes that failover stops streaming and starts polling, so it's not a generic failover of 2 options but a specific one from streaming to polling. Should we call this AdaptiveStreamingFetcher?

It then becomes unclear (if it's an AdaptiveStreamingFetcher) that it can be started without streaming... then failover just tries to start polling again (cause it starts with polling)

src/main/java/io/getunleash/repository/FeatureRepositoryImpl.java
gastonfournier: Based on my other comment I'm thinking should this be something like:
```suggestion
this.fetcher = config.isStreamingMode()? new AdaptiveFetcher(unleashConfig, featureBackupHandler, engine) : new PollingFeatureFetcher(
                        config,
                        config.getUnleashFeatureFetcherFactory().apply(config),
                        engine,
                        backup,
                        eventEmitter);
```
What I don't like about this option is that we're creating polling fetcher in two places, which makes it harder to maintain...

An alternative could be to make the `requestFailover` a no-op if the fetcher is already polling.

src/main/java/io/getunleash/repository/StreamingFeatureFetcherImpl.java
```diff
@@ -187,7 +189,12 @@ private void handleFailoverDecision(FailoverStrategy.FailEvent failEvent) {
                     "Streaming failover triggered: {}. Client is switching over to polling mode.",
                     failEvent.getMessage());
 
-            // changeToPolling()
+            if (modeController != null) {
+                modeController.requestFailover();
+            } else {
+                LOGGER.warn(
+                        "No ModeController configured, cannot request failover to polling mode.");
+            }
```
gastonfournier, line 197: I think this is what's causing the chicken and egg problem; maybe it's not the streaming fetcher who needs to control the streaming fetcher, but the streaming fetcher handling the failover means it needs to know the failover mechanism or know someone who knows... I think this knowledge is breaking encapsulation and should be refactored (but it's not this PR's fault)

chriswk (approved):
Having gone over this again, and having discussed with you off-github, no protests from me. Have a 👍

src/test/java/io/getunleash/repository/AdaptiveFetcherTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+        verify(streamingFeatureFetcher).start();
+        verify(pollingFeatureFetcher, never()).start();
+
+        adaptiveFetcher.requestFailover();
+
+        verify(streamingFeatureFetcher).stop();
+        verify(pollingFeatureFetcher).start();
+    }
+}
```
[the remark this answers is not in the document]
chriswk: Though I do think CoPilot has a point here.

src/main/java/io/getunleash/repository/StreamingFeatureFetcherImpl.java
sighphyre: I think there's always going to be some level of chicken and egg here. The only thing that has the information needed to decide if streaming is borked is streaming itself. That's current calculated in the failover, but the source of truth is still streaming. I'm open to suggestions here if you want to take a stab at it

PR #349, opened 2026-01-08
Minor tweaks to comment on

gastonfournier:


src/main/java/io/getunleash/repository/StreamingFeatureFetcherImpl.java
```diff
@@ -29,10 +30,10 @@ class StreamingFeatureFetcherImpl implements FetchWorker {
     private final UnleashEngine engine;
     private final BackupHandler featureBackupHandler;
     private final FailoverStrategy failoverStrategy;
-    private final ModeController modeController;
+    @Nullable private final ModeController modeController;
```
gastonfournier: This is only nullable for tests, maybe we can ignore this, but also simplify this class by not checking if modeController is null

src/main/java/io/getunleash/repository/StreamingFeatureFetcherImpl.java
```diff
@@ -55,7 +56,7 @@ class StreamingFeatureFetcherImpl implements FetchWorker {
             UnleashEngine engine,
             BackupHandler featureBackupHandler,
             FailoverStrategy failoverStrategy,
-            ModeController modeController) {
+            @Nullable ModeController modeController) {
```
gastonfournier: Same as before, this is only nullable for testing, but we're not testing a realistic scenario

src/main/java/io/getunleash/repository/StreamingFeatureFetcherImpl.java
```diff
@@ -176,20 +177,16 @@ private FailoverStrategy.FailEvent toFailEvent(Throwable throwable) {
...
     private void handleFailoverDecision(FailoverStrategy.FailEvent failEvent) {
-        boolean shouldFail = failoverStrategy.shouldFailover(failEvent, Instant.now());
-        if (shouldFail) {
-            LOGGER.warn(
-                    "Streaming failover triggered: {}. Client is switching over to polling mode.",
-                    failEvent.getMessage());
-
+        if (failoverStrategy.shouldFailover(failEvent, Instant.now())) {
             if (modeController != null) {
+                LOGGER.warn("Requesting streaming failover after: {}", failEvent.getMessage());
```
gastonfournier: I think this log makes more sense here, but as I said, maybe we don't need to check modeController != null... also just log that we're failing over, but this class doesn't have to know we're falling back to polling

PR #351, opened 2026-01-12
chore: clean up impact metrics public api

sighphyre:
This closes off most of the impact metrics API. None of this is necessary for the feature to work because the intended public surface is contained within `MetricsAPI`. 

I've had to rewrite a misbehaving test and move the histogram serializer to close off the histogram code. No logic changes but more significant than everything else here which is basically just remove the public keyword

### DIFF
--- a/src/main/java/io/getunleash/metric/DefaultHttpMetricsSender.java
+++ b/src/main/java/io/getunleash/metric/DefaultHttpMetricsSender.java
@@ -40,9 +40,7 @@ public class DefaultHttpMetricsSender implements MetricSender {
                         .registerTypeAdapter(Instant.class, new InstantSerializer())
                         .registerTypeAdapter(AtomicLong.class, new AtomicLongSerializer());
 
-        HistogramBucketSerializer.Register(gsonBuilder);
-
-        this.gson = gsonBuilder.create();
+        this.gson = HistogramBucketSerializer.Register(gsonBuilder).create();
     }
 
     public int registerClient(ClientRegistration registration) {
--- a/src/main/java/io/getunleash/metric/OkHttpMetricsSender.java
+++ b/src/main/java/io/getunleash/metric/OkHttpMetricsSender.java
@@ -60,9 +60,7 @@ public class OkHttpMetricsSender implements MetricSender {
                         .registerTypeAdapter(LocalDateTime.class, new DateTimeSerializer())
                         .registerTypeAdapter(AtomicLong.class, new AtomicLongSerializer());
 
-        HistogramBucketSerializer.Register(gsonBuilder);
-
-        this.gson = gsonBuilder.create();
+        this.gson = HistogramBucketSerializer.Register(gsonBuilder).create();
     }
 
     @Override


src/main/java/io/getunleash/impactmetrics/HistogramBucketSerializer.java
```diff
@@ -1,10 +1,10 @@
-package io.getunleash.util;
+package io.getunleash.impactmetrics;
```
sighphyre, line 1: Most of the other serializers live in `util` (where good code structure goes to die), but those serializers operate on built in types. This move is so that this serializer can peek into internal functions. The alternative is a a public class which I'm trying to avoid

src/main/java/io/getunleash/impactmetrics/HistogramBucketSerializer.java
```diff
@@ -21,4 +21,9 @@ public JsonElement serialize(
         jsonObject.addProperty("count", src.getCount());
         return jsonObject;
     }
+
+    public static GsonBuilder Register(GsonBuilder gsonBuilder) {
```
sighphyre, line 25: To avoid having to declare the `HistogramBucket` public, we expose this, which allows us to indirect back to this class while keeping the `HistrogramBucket` package private. Means this class needs to be public or this needs another home.

Don't see a better home for this and this is the kind of class I'm okay with being public so... *shrug*

src/main/java/io/getunleash/repository/StreamingFeatureFetcherImpl.java
sighphyre, line 80: Nothing in this class actually changed. Last PR that landed on main didn't get spotless applied so this is spotless arriving late to the party

src/test/java/io/getunleash/metric/DefaultHttpMetricsSenderTest.java
```diff
@@ -126,19 +127,19 @@ public void should_send_impact_metrics_with_histogram_and_plus_inf_bucket()
         UnleashConfig config = UnleashConfig.builder().appName("test-app").unleashAPI(uri).build();
 
         InMemoryMetricRegistry registry = new InMemoryMetricRegistry();
-        Histogram histogram =
-                registry.histogram(
-                        new BucketMetricOptions(
-                                "test_histogram", "testing histogram", List.of(1.0, 5.0)));
+        MetricsAPI metricsAPI =
```
sighphyre, line 130: Smells like an integration test but it mucks with internals to get its job done, which means making Histograms public.

Instead of that, we can just write a proper integration test and go via the intended public API and keep Histograms private

src/main/java/io/getunleash/metric/DefaultHttpMetricsSender.java
```diff
@@ -35,13 +34,15 @@ public DefaultHttpMetricsSender(UnleashConfig unleashConfig) {
...
                         .registerTypeAdapter(LocalDateTime.class, new DateTimeSerializer())
                         .registerTypeAdapter(Instant.class, new InstantSerializer())
-                        .registerTypeAdapter(AtomicLong.class, new AtomicLongSerializer())
-                        .registerTypeAdapter(HistogramBucket.class, new HistogramBucketSerializer())
-                        .create();
+                        .registerTypeAdapter(AtomicLong.class, new AtomicLongSerializer());
+
+        HistogramBucketSerializer.Register(gsonBuilder);
+
+        this.gson = gsonBuilder.create();
```
gastonfournier: Stylish suggestion, although yours looks fine, this feels more natural to me ("as if" the Register method was immutable, although it's not)
```suggestion
        this.gson = HistogramBucketSerializer.Register(gsonBuilder).create();
```

sighphyre: Sure, seems fair, the Register is builder-esque anyway, may as well it in this way. Done!

PR #352, opened 2026-01-13
fix: impact metrics now correctly resolves environment

sighphyre:
Bugfix to impact metrics. Environment should be resolved from the API key in the same way that it's being resolved in Node. Currently implementation resolves this from the config property "environment", which does not match the actual environment and defaults to "default" - so more or less always wrong.

I've always swapped out the `StaticContext` class for a `ImpactMetricsContext` class. StaticContext is a NodeSDK abstraction that the JavaSDK does not have, there's no point in adding it for this problem

### DIFF
--- a/src/main/java/io/getunleash/DefaultUnleash.java
+++ b/src/main/java/io/getunleash/DefaultUnleash.java
@@ -35,7 +35,7 @@ public class DefaultUnleash implements Unleash {
     private final UnleashContextProvider contextProvider;
     private final EventDispatcher eventDispatcher;
     private final UnleashConfig config;
-    private final MetricsAPIImpl impactMetrics;
+    private final MetricsAPI impactMetrics;
 
     private static EngineProxy defaultToggleRepository(
             UnleashConfig unleashConfig, Strategy... strategies) {


src/main/java/io/getunleash/impactmetrics/ImpactMetricContext.java
```diff
@@ -0,0 +1,31 @@
+package io.getunleash.impactmetrics;
+
+import io.getunleash.util.UnleashConfig;
+
+public class ImpactMetricContext {
```
sighphyre, line 5: This more or less just renames StaticContext to this and patches the getEnvironment method. 

Currently public, not sure it needs to be, except that making this non public requires some surgery to the MetricsAPI constructor, which I don't want to tinker with right now. I think this kinda safe enough. Open to disagreement

src/main/java/io/getunleash/impactmetrics/MetricFlagContext.java
```diff
@@ -3,7 +3,7 @@
 import io.getunleash.UnleashContext;
 import java.util.List;
 
-class MetricFlagContext {
+public class MetricFlagContext {
```
sighphyre, line 6: Yeah so this was actually in the public API, and needs to be a public type

src/main/java/io/getunleash/impactmetrics/MetricsAPI.java
```diff
@@ -1,141 +1,31 @@
 package io.getunleash.impactmetrics;
```
sighphyre, line 2: Turned this into an interface but its a public dependency and exposing this to the public API as a concrete type is Not A Good Idea™

src/main/java/io/getunleash/impactmetrics/MetricsAPIImpl.java
```diff
@@ -0,0 +1,152 @@
+package io.getunleash.impactmetrics;
```
sighphyre, line 1: Unchanged, just a rename + inherit interface

src/main/java/io/getunleash/impactmetrics/ImpactMetricContext.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public ImpactMetricContext(UnleashConfig config) {
+        this.config = config;
+    }
+
+    public String getAppName() {
+        return config.getAppName();
+    }
+
+    private static String extractEnvironment(String apiKey) {
```
sighphyre, line 16: This is basically the fix - we figure out the environment from the apiKey specified in the config. The config determines the API key from either the headers or header resolver so it should always match the active API key

src/main/java/io/getunleash/FakeUnleash.java
```diff
@@ -32,10 +29,46 @@ public class FakeUnleash implements Unleash {
     private final MetricsAPI impactMetrics;
 
     public FakeUnleash() {
-        ImpactMetricRegistry registry = new InMemoryMetricRegistry();
-        VariantResolver variantResolver = (flagName, context) -> Variant.DISABLED_VARIANT;
-        StaticContext staticContext = new StaticContext("fake-app", "test");
-        this.impactMetrics = new MetricsAPI(registry, variantResolver, staticContext);
+        this.impactMetrics =
```
sighphyre, line 32: Put in a no-op interface by default. The concrete type is not useful here, since it only accepts data and cannot be queried. This at least gives us a chance for the end user to swap something in here if they need to

src/main/java/io/getunleash/FakeUnleash.java
```diff
@@ -32,10 +29,46 @@ public class FakeUnleash implements Unleash {
...
+                    @Override
+                    public void observeHistogram(String name, double value) {}
+
+                    @Override
+                    public void observeHistogram(
+                            String name, double value, MetricFlagContext ctx) {}
+                };
+    }
+
+    public FakeUnleash(MetricsAPI metricsAPI) {
```
sighphyre, line 70: If you want to muck with impact metrics as part of your test scaffold, now you can

src/test/java/io/getunleash/impactmetrics/ImpactMetricsContextTest.java
```diff
@@ -0,0 +1,41 @@
+package io.getunleash.impactmetrics;
```
sighphyre, line 1: New tests

src/main/java/io/getunleash/impactmetrics/ImpactMetricContext.java
gastonfournier: Agree, all data from this class is already accessible publicly in the SDK or user-provided

src/main/java/io/getunleash/impactmetrics/ImpactMetricContext.java
gastonfournier: I think development is a safe default, especially for this new feature which implies having a newer version of unleash-server

src/main/java/io/getunleash/DefaultUnleash.java
```diff
@@ -34,7 +35,7 @@ public class DefaultUnleash implements Unleash {
     private final UnleashContextProvider contextProvider;
     private final EventDispatcher eventDispatcher;
     private final UnleashConfig config;
-    private final MetricsAPI impactMetrics;
+    private final MetricsAPIImpl impactMetrics;
```
gastonfournier:
```suggestion
    private final MetricsAPI impactMetrics;
```
Should use the interface here, but it will also need to be imported

gastonfournier (approved):
LGTM, just a minor comment on using interface instead of actual implementation

src/main/java/io/getunleash/DefaultUnleash.java
sighphyre: Yep, fair enough, patched

PR #359, opened 2026-01-21
docs: improve impact metrics description

kwasniew:
## About the changes

* show more realistic gauge example that is correct when multiple pods are used
* make histograms example use explicit buckets

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,30 +351,24 @@ unleash.getImpactMetrics().incrementCounter("request_count");
 
 ### Gauges
 
-Use gauges for values that can go up and down, such as current memory usage or active thread count.
+Use gauges for point-in-time values that can go up or down.
 
 ```java
 unleash.getImpactMetrics()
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -351,30 +351,24 @@ unleash.getImpactMetrics().incrementCounter("request_count");
...
 
-long currentHeap = Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory();
-
-unleash.getImpactMetrics().updateGauge("heap_memory_total", currentHeap);
+unleash.getImpactMetrics().updateGauge("total_users", userCount);
 ```
 
 ### Histograms
 
-Use histograms to measure the distribution of values, such as request duration or response size. Unleash automatically calculates percentiles (p50, p95, p99).
```
kwasniew, line 367: we don't explain aggregation fns for other metrics and this info is part of the server side logic so I removed p50 and others

PR #360, opened 2026-01-21
fix: change gauge from long to double

sjaanus:
Need to allow decimal numbers of gauge values.

Since we have `NumericMetricSample` shared with counter and gauge, we need to do some casting for counter.


src/main/java/io/getunleash/impactmetrics/CounterImpl.java
```diff
@@ -38,12 +38,12 @@ CollectedMetric collect() {
         for (String key : values.keySet()) {
             Long value = values.remove(key);
             if (value != null) {
-                samples.add(new NumericMetricSample(parseLabelKey(key), value));
+                samples.add(new NumericMetricSample(parseLabelKey(key), value.doubleValue()));
```
sjaanus: Need to do this casting due to shared `NumericMetricSample`, but I think it is okay tradeoff.

kwasniew (commented):
I had the same issue with casting in Rust. I prefer to have two separate types instead

PR #378, opened 2026-05-22
task: Use getInputStream to get data

### LINKED ISSUE #377: HttpFeatureFetcher with GraalVM

Hi,
I am using the unleash sdk in a quarkus project run as a native image using graalvm (jdk 21). When not doing any special configurations, the HttpFeatureFetcher fails with an error `java.net.ContentHandlerFactory: Provider sun.awt.[www.content.MultimediaContentHandlers](https://www.content.multimediacontenthandlers/) not found` when the fetcher does `request.getContent()` [here](https://github.com/Unleash/unleash-java-sdk/blob/01f1aa226e8034a1ffa4d2119ad5efd2f8f87580/src/main/java/io/getunleash/repository/HttpFeatureFetcher.java#L77). This is not an issue of the sdk itself, more of how graalvm works and what it includes in native builds, but I was wondering if it wouldn't be better in HttpFeatureFetcher to use `request.getInputStream()` as that would be more efficient and less error prone than `(InputStream)request.getContent()`, and as a happy side-effect would remove the issue in the native image as it would no longer unnecessarily try to to find content handlers via service loader.
For now I am using the OkHttpFeatureFetcher instead, but would like it if I wouldn't have to include the extra dependency on okhttp (which only works with okhttp 4.x it seems, which is also rather outdated).

### Background

_No response_

### Solution suggestions

_No response_

chriswk:
As reported in #377 - getContent() uses serviceloaders to locate the correct content processor. With GraalVM, depending on your configuration it can locate the wrong content processor and give you bad data.

 We've already checked that we don't have a gzip stream, so we're trusting our upstream to return application/json and do not need a content processor. This saves us a cast from content to InputStream as well.

fixes: #377

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <version.unleash.specification>5.2.2</version.unleash.specification>
-        <arguments />
+        <arguments/>
         <version.assertj>3.27.7</version.assertj>
         <version.gson>2.13.2</version.gson>
         <version.jackson>2.20.1</version.jackson>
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
This looks good

PR #381, opened 2026-06-16
chore: added an sbom task for attaching sboms to github releases

chriswk:
This adds the cyclonedx plugin to our pom file, and sets up a task for generating and attaching sboms to the github releases we make.


sastromskis (approved):
NICE!

PR #385, opened 2026-08-24
chore: bump ygg dependency to 1.0.3

daveleek:
This bumps the dependency on the Unleash Yggdrasil engine for java to 1.0.3, making Java SDK conform to the way SemVer metadata should work


pom.xml
```diff
@@ -61,7 +61,7 @@
         <dependency>
             <groupId>io.getunleash</groupId>
             <artifactId>yggdrasil-engine</artifactId>
-            <version>0.5.1</version>
+            <version>1.0.3</version>
```
gastonfournier, line 64: Is this a big bump?

gastonfournier (approved):
LG, just wondering if this needs a major... if it's only semver I'd say no

pom.xml
daveleek: 1.0.0 is CIDR support and SemVer if I read things correctly. Resulted in a minor version bump on the engines, so maybe a minor bump here as well when we do the release?

That's separate from this PR tho so can be decided when doing the release

daveleek: This was [a minor bump in .NET](https://github.com/Unleash/unleash-dotnet-sdk/releases/tag/6.2.0) so I'll do a minor bump when releasing this

